PR #2, opened 2019-01-25
rails 5.1 support

kbrock:
Move our rails 5.1 `virtual_attribute` changes from manageiq to here.

NOTE:

tested:
- `join_keys` change
- `aliases` => `_`

untested:
- `&block`
- `find_with_associations`

see also https://github.com/ManageIQ/manageiq/pull/18076

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -3,14 +3,16 @@
   appraise "#{db_gem}-#{ar_version.split('.').first(2).join}" do
     gem "activerecord", "~> #{ar_version}"
 
-    if ar_version >= "5.1"
-      gem "pg"
+    gem "pg"
+    if ar_version >= "5.0"
       gem "mysql2"
+    else
+      gem "mysql2", '~> 0.4.0'
+    end
+    if ar_version >= "5.2"
       gem "sqlite3"
     else
       gem "sqlite3", "~> 1.3.6"
-      gem "pg", " ~> 0.18.4"
-      gem "mysql2", '~> 0.4.0'
     end
   end
 end
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -176,7 +176,7 @@ module ActiveRecord
       end
     end
 
-    include Module.new {
+    include(Module.new {
       # From ActiveRecord::FinderMethods
       def find_with_associations
         real = without_virtual_includes
@@ -221,6 +221,6 @@ module ActiveRecord
 
         real.calculate(operation, attribute_name)
       end
-    }
+    })
   end
 end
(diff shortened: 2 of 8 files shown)


kbrock:
@juliancheal and @jrafanie These changes look like the changes we needed to make in the miq repo?

Appraisals
```diff
@@ -3,14 +3,16 @@
   appraise "#{db_gem}-#{ar_version.split('.').first(2).join}" do
     gem "activerecord", "~> #{ar_version}"
 
-    if ar_version >= "5.1"
-      gem "pg"
+    gem "pg" #, " ~> 0.18.4"
```
NickLaMuro: Pedantic:  Remove the comment?

juliancheal:
giphy

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -176,15 +176,22 @@ def without_virtual_includes
...
+        if ActiveRecord.version.to_s >= "5.1"
+          recs, join_dep = real.find_with_associations { |relation, join_dependency| [relation, join_dependency] }
+        else
+          recs = real.find_with_associations
+        end
         MiqPreloader.preload(recs, preload_values + includes_values) if includes_values
 
+        if ActiveRecord.version.to_s >= "5.1"
+          yield recs, join_dep
+        end
```
jrafanie: Can these conditionals be removed?  I thought 5.0 and earlier didn't take advantage of them.  Maybe yield if block_given?  I was hoping we can do the logic for 5.1 in a 5.0 compatible way.

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -110,7 +110,11 @@ def virtual_aggregate_arel(reflection, method_name, column)
 
           foreign_table = reflection.klass.arel_table
           # need db access for the keys, so delaying all this lookup until call time
-          join_keys = reflection.join_keys(reflection.klass)
+          if ActiveRecord.version.to_s >= "5.1"
```
jrafanie: I don't know how to do this without the version check.  So I'm 👍 on this.

lib/active_record/virtual_attributes/virtual_fields.rb
kbrock: `block_given?` instead?

kbrock: part of the version check tells us how to proceed further when we deprecate 5.0
So we'll know to remove the extra if check.

While version checking is uglier, it does tell us how to deprecate code for older versions

jrafanie: Right, I'm not sure if it's even needed though.  I'm fine if it is.

jrafanie: Looks good 👍

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -176,15 +176,23 @@ def without_virtual_includes
...
+        else
+          recs = real.find_with_associations
+        end
         MiqPreloader.preload(recs, preload_values + includes_values) if includes_values
 
+        # when 5.0 support is dropped, assume a block given
+        if block_given?
+          yield recs, join_dep
+        end
         recs
```
jrafanie: 🤣 ^  oops

After much debugging @kbrock and I found it should have been:

```ruby
        if block_given?
          yield recs, join_dep
        else
          recs
        end
```

PR #8, opened 2019-04-08
Remove COUNT(*) ORDER BY

kbrock:
postgres does not allow a `COUNT(*)` to have an `ORDER BY` in the query.

This removes the `ORDER BY` from `COUNT(*)` queries. (all virtual_total queries are aggregate queries, so all of them do not want the order.)

NOTE: in mysql and sqlite, nulls sort before other value, but in postgres, they come later.
One of the tests sorted with a null and would fail in postgres. `author0` had no books.

I changed a test to not deal with this edge case.
Will deal with this when we get virtual aggregates fixed in #15

### DIFF
--- a/spec/virtual_total_spec.rb
+++ b/spec/virtual_total_spec.rb
@@ -114,7 +114,6 @@ describe VirtualAttributes::VirtualTotal do
 
     context "with a has_many that includes an order" do
       it "sorts by total" do
-        skip("fix order in scopes") if ENV["DB"] == "pg"
         author2 = Author.create_with_books(2)
         author2.create_books(2, :published => true, :rating => 5)
         author0 = Author.create
@@ -143,8 +142,6 @@ describe VirtualAttributes::VirtualTotal do
       end
 
       it "can bring back totals in primary query" do
-        skip("fix order in scopes") if ENV["DB"] == "pg"
-
         author3 = Author.create_with_books(3)
         author3.create_books(2, :published => true, :rating => 2)
         author1 = Author.create_with_books(1)
@@ -349,7 +346,6 @@ describe VirtualAttributes::VirtualTotal do
       end
 
       it "uses calculated (inline) attribute" do
-        skip("fix order in scopes") if ENV["DB"] == "pg"
         auth1 = model_with_children(0)
         auth2 = model_with_children(2)
         query = base_model.select(:id, :total_ordered_books).load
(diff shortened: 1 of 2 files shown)


kbrock:
@NickLaMuro pointed me to the original code that added this: https://github.com/ManageIQ/manageiq/pull/16518

spec/virtual_total_spec.rb
```diff
@@ -88,12 +88,12 @@
 
     context "with order clauses in the relation" do
       it "sorts by total" do
-        skip("fix order in scopes") if ENV["DB"] == "pg"
```
NickLaMuro: I guess my only concern with this change is with these specs.

I don't have the full context, but were these working on other DB's, but then now aren't with PG?  Are we only testing against SQLite and PG?  If so, then was SQLite the one that was allowing the `ORDER BY`?  But then why did we change the values below in the `.create_books` calls?

kbrock: good questions.

1. in miq, we were only testing sqlite
2. in the gem, we started testing mysql and sqlite3 for this test (pg for others too)
3. with this change, we are not testing for all 3

create_books changed because in pg, with sorting, nulls are last, but in sqlite and mysql null is first. The change over there removed the null value so we could have an actual number there, and we'd avoid the nuance with null last (I was trying to focus on not blowing up. the nuance of a null was beyond the scope of the test)

NickLaMuro: > create_books changed because in pg, with sorting, nulls are last, but in sqlite and mysql null is first.

So, the first arg in `create_books` (since [I wrote it](https://github.com/NickLaMuro/manageiq/commit/f8dddabc)) is just changing the number of books being created.  So I guess with how this was done previously (since every author seemed to have at least 1 published book), I don't see why it is necessary to up the counts.

Am I missing something?

kbrock: ok. removed those extra changes, just added the `author1.create_books` to avoid the null

kbrock:
status: this is done
put a wip in place for getting a sum of no records to return a 0 - then this pr will merge without funkiness added to our specs

(alternatively, we could merge this with a tweak, and in the other PR we could undo this change)

spec/virtual_total_spec.rb
kbrock: I had upped the count so there were no authors with no published books.
it made the sorting more consistent.

since then, I've put in a pr so it always returns a 0, so this is not an issue

kbrock:
update: this is failing because postgres defaults to `NULLS LAST` and mysql/sqlite defaults to `NULLS FIRST`.

the blocked PR removes the nulls from this attribute so sorting works

kbrock:
un-WIP - just changing the test. need to get this fix in and can't wait for the other change

jrafanie:
LGTM, will wait until we decide what we do about code climate

jrafanie:
Merging, code climate error is unrelated and this code change looks fine to me.

PR #9, opened 2019-04-08
code climate tweaks

kbrock:
These are code formatting only to fix code climate errors

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -58,7 +58,7 @@ module ActiveRecord
 
   module Associations
     class Preloader
-      prepend Module.new {
+      prepend(Module.new {
         def preloaders_for_one(association, records, scope)
           klass_map = records.compact.group_by(&:class)
 
@@ -67,14 +67,14 @@ module ActiveRecord
             preload(subset, virtuals)
           end
 
-          records_with_association = klass_map.select { |k, rs| k.reflect_on_association(association) }.flat_map { |k, rs| rs }
+          records_with_association = klass_map.select { |k, _rs| k.reflect_on_association(association) }.flat_map { |_k, rs| rs }
           if records_with_association.any?
             loaders.concat(super(association, records_with_association, scope))
           end
 
           loaders
         end
-      }
+      })
     end
 
     # FIXME: Hopefully we can get this into Rails core so this is no longer


lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -58,7 +58,7 @@ class Base
 
   module Associations
     class Preloader
-      prepend Module.new {
+      prepend(Module.new {
```
NickLaMuro: Ewww... can we instead tell `rubocop` to not care about this?  I think it looks worse this way...

That, or maybe make this a named module above instead of using `Module.new`?


```ruby
module ActiveRecord
  module VirtualAttributes
    module VirtualFields
      PreloaderForOnePatch
    end
  end

  module Associations
    class Preloader
      prepend VirtualAttributes::VirtualFields:PreloaderForOnePatch
  end
end
```

NickLaMuro: Also noticed while doing the above, you re-open the `ActiveRecord` module on line `54` when you were already doing so at the top of the file.

Fryguy: This can actually be quite important, and I think rubocop is right.  The problem stems from the binding of the block.  This is one of the drawback of the seattle style, in that it's not clear if the block binds prepend with Module.new as a parameter or if the block binds to Module.new and the combination is the parameter.  Even more confusing is that `{}` binds differently than `do; end`.  This parens ensure that from a readability standpoint you know what is binding where.

All that said, I 100% agree with you about anonymous modules vs defined, and I think defined is *much* clearer and clears up this ambiguity.  Note that the seattle proponents argue this point a a counter to the above; that is, if the code is ambiguous without parens, then you should change the code because it probably wasn't clear *with* the parens either.

Fryguy: @NickLaMuro I opened https://github.com/ManageIQ/activerecord-virtual_attributes/issues/11 in response to your second comment, and based on organization discussion @kbrock and I had.

Fryguy: @kbrock I think with the parens you can also now switch to do; end style (up to you)

NickLaMuro: > This can actually be quite important, and I think rubocop is right. The problem stems from the binding of the block...

@Fryguy That is fine, I just cringe the minute I start seeing `something({  ... })` because I start thinking I am in `C` land... or worse... `javascript`...  But yeah, I think the defined module makes more sense long term, since this is now in it's own code base, and we are treating this as an extension of the `ActiveRecord` API.

But I think merging now and revisiting makes sense, since that was kind of my intent.  👍

PR #10, opened 2019-04-09
 Fix includes(va).references(va) in rails 5.1

kbrock:
In rails 5.1, it was not properly aliasing the names of columns
after an `includes().references()`

The attributes were coming across as the alias name instead of the attribute name

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -195,8 +195,9 @@ module ActiveRecord
         # when 5.0 support is dropped, assume a block given
         if block_given?
           yield recs, join_dep
+        else
+          recs
         end
-        recs
       end
 
       # From ActiveRecord::QueryMethods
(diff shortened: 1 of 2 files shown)


spec/virtual_attributes_spec.rb
```diff
@@ -983,4 +983,13 @@ def col2
     tc.save
     expect(tc.attributes.size).to eq(2)
   end
+
+  it "doesn't botch up the attributes when requiring", :with_test_class do
```
jrafanie: referencing?  I'm not sure what you mean by requiring

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -187,13 +187,17 @@ def find_with_associations
         else
           recs = real.find_with_associations
         end
-        MiqPreloader.preload(recs, preload_values + includes_values) if includes_values
+
+        if includes_values
+          ActiveRecord::Associations::Preloader.new.preload(recs, preload_values + includes_values)
```
Fryguy: :tada:  Glad you got rid of MiqPreloader here 😄

spec/virtual_attributes_spec.rb
kbrock: stupid `s/require/references/g`

kbrock:
fixed spec text. let me know if there is anything else

PR #12, opened 2019-04-10
Support legacy types for attributes

kbrock:
This PR adds support for the data types used by previous versions of this extension.f

Some parts of the app are throwing errors when deserializing yaml from the database:

```
ArgumentError: undefined class/module VirtualAttributes::Type::
```

This happens because a data type was stored in yaml. Those values have changed and the existing one can't be found anymore

/cc @mzazrivec This should solve your problem

### DIFF
--- a/lib/active_record/virtual_attributes.rb
+++ b/lib/active_record/virtual_attributes.rb
@@ -141,3 +141,12 @@ end
 
 require "active_record/virtual_attributes/virtual_total"
 require "active_record/virtual_attributes/arel_groups"
+
+# legacy support for sql types
+module VirtualAttributes
+  module Type
+    Symbol      = ActiveRecord::VirtualAttributes::Type::Symbol
+    StringSet   = ActiveRecord::VirtualAttributes::Type::StringSet
+    NumericSet  = ActiveRecord::VirtualAttributes::Type::NumericSet
+  end
+end


djberg96:
I fully approve of this branch name.

lib/active_record/virtual_attributes.rb
```diff
@@ -141,3 +141,8 @@ def with_value_from_database(value)
 
 require "active_record/virtual_attributes/virtual_total"
 require "active_record/virtual_attributes/arel_groups"
+
+# legacy support for sql types
```
Fryguy: Do we need to wrap this in some Rails version?  (Not sure what the word "legacy" refers to here)

himdel:
not actually defining `VirtualAttributes::Type` anywhere...

lib/active_record/virtual_attributes.rb
kbrock: miq was storing this type name in our database
So legacy means miq databases

kbrock:
> not actually defining VirtualAttributes::Type anywhere...
> - @himdel 

Are you suggesting adding that module, then adding those types inside it?

himdel:
> Are you suggesting adding that module, then adding those types inside it?

Pretty much, yeah.. travis is also still complaining with `NameError: uninitialized constant VirtualAttributes::Type`.

kbrock:
@himdel does this work better for you?

himdel:
Well, the tests went green :+1: 
Looks good, but I didn't have that problem in the first place..

mzazrivec:
I can confirm, that with these changes in place, the error no longer shows.

kbrock:
@Fryguy you ok with this change - people's databases are failing

mzazrivec:
@Fryguy ping? Any take on this? This problem causes notification errors in current master UI and it needs to be resolved.

lib/active_record/virtual_attributes.rb
jrafanie: We can clean up the message later, I think this is support for pre-extraction virtual attributes

PR #13, opened 2019-04-15
[WIP] Null when no associated records (fix sorting)

kbrock:
extracted:

- [x] #14 extracting tests - lots of the things floating around assume the tests are there.

for attributes with null values, doesn't go back to the database multiple times

for aggregates (currently only really use `sum`) - it returns 0

so when you sort by aggregates, the 0's (which are null behind the sceens) sort properly

this fixes an issue with #8

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_total.rb
+++ b/lib/active_record/virtual_attributes/virtual_total.rb
@@ -81,11 +81,10 @@ module VirtualAttributes
             else
               rel = send(relation)
               if rel.loaded?
-                rel.blank? ? nil : rel.map { |t| t.send(column) }.compact.send(method_name)
+                rel_values = rel.map { |t| t.send(column) }.compact
+                rel_values.blank? ? nil : rel_values.send(method_name)
               else
-                # virtual attributes support for aggregates
-                arel_column = rel.klass.arel_attribute(column)
-                rel.try(method_name, arel_column) || 0
+                rel.try(method_name, rel.klass.arel_attribute(column))
               end
             end
           end
--- a/lib/active_record/virtual_attributes/virtual_total.rb
+++ b/lib/active_record/virtual_attributes/virtual_total.rb
@@ -51,7 +51,7 @@ module VirtualAttributes
       #    # arel => (SELECT sum("disks"."size") where "hardware"."id" = "disks"."hardware_id")
 
       def virtual_aggregate(name, relation, method_name = :sum, column = nil, options = {})
-        define_virtual_aggregate_method(name, relation, method_name, column)
+        define_virtual_method(name, relation, method_name, column)
         reflection = reflect_on_association(relation)
 
         if options.key?(:arel)
@@ -69,23 +69,31 @@ module VirtualAttributes
         end
       end
 
-      def define_virtual_aggregate_method(name, relation, method_name, column)
+      def define_virtual_method(name, relation, method_name, column)
         if method_name == :size
-          define_method(name) do
-            has_attribute?(name) ? self[name] : (send(relation).try(:size) || 0)
-          end
+          define_virtual_size_method(name, relation)
         else
-          define_method(name) do
-            if has_attribute?(name)
-              self[name]
+          define_virtual_aggregate_method(name, relation, method_name, column)
+        end
+      end
+
+      def define_virtual_size_method(name, relation)
+        define_method(name) do
+          has_attribute?(name) ? self[name] : (send(relation).try(:size) || 0)
+        end
+      end
+
+      def define_virtual_aggregate_method(name, relation, method_name, column)
+        define_method(name) do
+          if has_attribute?(name)
+            self[name]
+          else
+            rel = send(relation)
+            if rel.loaded?
+              rel_values = rel.map { |t| t.send(column) }.compact
+              rel_values.blank? ? nil : rel_values.send(method_name)
             else
-              rel = send(relation)
-              if rel.loaded?
-                rel_values = rel.map { |t| t.send(column) }.compact
-                rel_values.blank? ? nil : rel_values.send(method_name)
-              else
-                rel.try(method_name, rel.klass.arel_attribute(column))
-              end
+              rel.try(method_name, rel.klass.arel_attribute(column))
             end
           end
         end
--- a/lib/active_record/virtual_attributes/virtual_total.rb
+++ b/lib/active_record/virtual_attributes/virtual_total.rb
@@ -72,21 +72,22 @@ module VirtualAttributes
       def define_virtual_aggregate_method(name, relation, method_name, column)
         if method_name == :size
           define_method(name) do
-            (attribute_present?(name) ? self[name] : nil) || send(relation).try(:size) || 0
+            has_attribute?(name) ? self[name] : (send(relation).try(:size) || 0)
           end
         else
           define_method(name) do
-            (attribute_present?(name) ? self[name] : nil) ||
-              begin
-                rel = send(relation)
-                if rel.loaded?
-                  rel.blank? ? nil : rel.map { |t| t.send(column) }.compact.send(method_name)
-                else
-                  # aggregates are not smart enough to handle virtual attributes
-                  arel_column = rel.klass.arel_attribute(column)
-                  rel.try(method_name, arel_column) || 0
-                end
+            if has_attribute?(name)
+              self[name]
+            else
+              rel = send(relation)
+              if rel.loaded?
+                rel.blank? ? nil : rel.map { |t| t.send(column) }.compact.send(method_name)
+              else
+                # virtual attributes support for aggregates
+                arel_column = rel.klass.arel_attribute(column)
+                rel.try(method_name, arel_column) || 0
               end
+            end
           end
         end
       end
--- a/lib/active_record/virtual_attributes/virtual_total.rb
+++ b/lib/active_record/virtual_attributes/virtual_total.rb
@@ -80,7 +80,7 @@ module VirtualAttributes
               begin
                 rel = send(relation)
                 if rel.loaded?
-                  rel.blank? ? nil : (rel.map { |t| t.send(column).to_i } || 0).send(method_name)
+                  rel.blank? ? nil : rel.map { |t| t.send(column) }.compact.send(method_name)
                 else
                   # aggregates are not smart enough to handle virtual attributes
                   arel_column = rel.klass.arel_attribute(column)
--- a/spec/virtual_total_spec.rb
+++ b/spec/virtual_total_spec.rb
@@ -405,15 +405,15 @@ describe VirtualAttributes::VirtualTotal do
         authors
         query = Author.select(:id, :sum_recently_published_books_rating).order(:id).load
         expect do
-          expect(query.map(&:sum_recently_published_books_rating)).to eq([6, 5, 0, 0])
-        end.to match_query_limit_of(2)
+          expect(query.map(&:sum_recently_published_books_rating)).to eq([6, 5, nil, nil])
+        end.to match_query_limit_of(0)
       end
 
       it "calculates sum from attribute (and preloaded association)" do
         authors
         query = Author.includes(:recently_published_books).select(:id, :sum_recently_published_books_rating).order(:id).load
         expect do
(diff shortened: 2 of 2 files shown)


lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -72,21 +72,22 @@ def virtual_aggregate(name, relation, method_name = :sum, column = nil, options
...
-                  # aggregates are not smart enough to handle virtual attributes
-                  arel_column = rel.klass.arel_attribute(column)
-                  rel.try(method_name, arel_column) || 0
-                end
+            if has_attribute?(name)
+              self[name] || 0
+            else
+              rel = send(relation)
+              if rel.loaded?
+                rel.map { |t| t.send(column) }.compact.send(method_name)
```
NickLaMuro: Being blunt:  The commit message for this change is kind of a disaster... for reference:

> ### simplify loaded rel sum
> 
> array.map {} || 0 :
> 
> if there is an array, then it will always have a value,
>   so || 0 is never going to have a value
> 
> if an array is empty, empty.map {} will just be an empty value
> [].sum == 0
> 
> removed to_i since we could sum up floats

```diff
- rel.blank? ? nil : (rel.map { |t| t.send(column).to_i } || 0).send(method_name)
+ rel.map { |t| t.send(column) }.compact.send(method_name)
```


First off, the `compact` was completely unmentioned, but that plays a role in handling `nil` values that existed in the array previously (as `nil.to_i == 0`).

The `[] || 0` is actually a problem (if it was ever reachable) since `0.sum` is not a method on integers.

Finally, you removed `nil` being a valid return value, and that used to not be the case.

Regarding the formatting of the commit message, I think you would be better itemizing the changes (something like this, and probably could be tweaked):

```
simplify loaded rel sum

- removes `|| 0` since `0.sum` is invalid and it is not reachable (since `[] || 0 == []`)
- `[].sum == 0`, so we can avoid needing a `... || 0` case
- removes `nil` as being a possible return value when array is empty (now returns `0`)
- call `.compact` to remove `nil` values fromarray to avoid issue with `.send` (aka: `.sum`) later
```

* * *

The other issue I have is that we are now potentially returning floats when previously we casted to integers, and if that might potentially cause an issue down the line.

spec/virtual_total_spec.rb
```diff
@@ -328,4 +345,45 @@ def model_with_children(count)
...
+        author.create_books(1, :published => true, :rating => 2)
+        author.create_books(1, :published => true)
+      end
+    end
+
+    it "calculates sum locally" do
+      author
+      expect do
+        expect(author.sum_recently_published_books_rating).to eq(6)
+      end.to match_query_limit_of(1)
```
NickLaMuro: Can we have the test data for this test make this a `match_query_limit_of(2)` instead of `1`.

The  "calculates sum in database" has less of an impact when this spec calls the same number of queries.

spec/virtual_total_spec.rb
```diff
@@ -328,4 +345,45 @@ def model_with_children(count)
...
+        author.create_books(1, :published => true, :rating => 4)
+        author.create_books(1, :published => true, :rating => 2)
+        author.create_books(1, :published => true)
+      end
+    end
+
+    it "calculates sum locally" do
+      author
+      expect do
+        expect(author.sum_recently_published_books_rating).to eq(6)
```
NickLaMuro: Aside:  This is a really dumb attribute... and I think I wrote it....

I feel that `avg_published_book_rating` would be something a little more realistic.

NickLaMuro:
@kbrock Knowing you, there is a important issue that I forgot to make more prominent in my review regarding the fact these totals now can be floats:

https://github.com/ManageIQ/activerecord-virtual_attributes/pull/13#discussion_r275924601

Please address that in the review response.

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -40,7 +40,7 @@ def virtual_total(name, relation, options = {})
       #
       #    def allocated_disk_storage
       #      if disks.loaded?
-      #        disks.blank? ? nil : disks.map { |t| t.size.to_i }.sum
+      #        disks.map { |t| t.size }.compact.sum
```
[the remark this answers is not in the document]
kbrock: thanks - I went ahead and removed that change.
it was a change, and this PR is more about trying to keep down those changes

spec/virtual_total_spec.rb
kbrock: lol. was TOTALLY seeing the same thing.

Once we introduce average (which we don't in our app) - then the `nil` case really starts to show the issue of the `compact`

lib/active_record/virtual_attributes/virtual_total.rb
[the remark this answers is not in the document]
kbrock: removed this change - so this is all set

lib/active_record/virtual_attributes/virtual_total.rb
kbrock: if the fields are string'd numbers, it worked before and does not work now. but not sure we should work around this. they screwed up.
if the fields were floats or bigints before, it would return a different value (or error) depending upon the source. Now it returns the correct value.

it still returns different values for unloaded vs loaded associations - 0 vs nil - this has not been addressed by this pr. Trying to keep this PR as a documentation/adding tests and less of a change.

spec/virtual_total_spec.rb
kbrock: averages would be great - really show the issues with the nils

But then I'd have to come up with a sum test - and I don't want to deal with that right now :(
I'll make this change if you come up with another sum attribute

lib/active_record/virtual_attributes/virtual_total.rb
kbrock: I changed the comments. will split apart that commit

kbrock:
I fixed the codeclimate issue (even though I didn't break it on this code change)

I added a test to show the 0 vs nil for another edge case.
all (or most?) of the edge cases are now tested, so we can see the inconsistency in returning a 0 vs nil

The next step is to introduce a default value, so we can decide if we want a 0 or a nil when the relationships have no values. I like the default of a 0, but we do have cases that we seem to want a nil to come back.

I'll need to add quite a few tests around this, so adding defaults will take a little more time

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -72,21 +72,22 @@ def virtual_aggregate(name, relation, method_name = :sum, column = nil, options
...
-              begin
-                rel = send(relation)
-                if rel.loaded?
-                  rel.blank? ? nil : (rel.map { |t| t.send(column).to_i } || 0).send(method_name)
-                else
-                  # aggregates are not smart enough to handle virtual attributes
-                  arel_column = rel.klass.arel_attribute(column)
-                  rel.try(method_name, arel_column) || 0
-                end
+            if has_attribute?(name)
```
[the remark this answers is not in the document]
kbrock: good point - done

spec/virtual_total_spec.rb
kbrock: @kbrock do you mean having 2 different authors? - so it shows the N+1?

spec/virtual_total_spec.rb
```diff
@@ -328,4 +361,65 @@ def model_with_children(count)
       end
     end
   end
+
+  describe ".virtual_aggregation" do
```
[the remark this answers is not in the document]
NickLaMuro: Upon further inspection, not really, since they are both part of the same file.

Just seems like `virtual_total` inherits from `virtual_aggregation` conceptually, but instead, it is the reverse.

For now, carry on.

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -72,21 +72,21 @@ def virtual_aggregate(name, relation, method_name = :sum, column = nil, options
...
-                end
-              end
+            return self[name] || 0 if has_attribute?(name)
+            rel = send(relation)
+            if !rel.loaded?
+              # virtual attributes support for aggregates
+              arel_column = rel.klass.arel_attribute(column)
+              rel.try(method_name, arel_column) || 0
+            elsif rel.blank?
+              nil
```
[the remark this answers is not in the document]
kbrock: @NickLaMuro yes, change in direction - pairing this one back

spec/virtual_total_spec.rb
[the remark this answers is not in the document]
kbrock: :( `virtual_total` calls `virtual_aggregate` - so yea, total should inherit from aggregation.

ideas how to make this more clear (probably a separate PR)

spec/virtual_total_spec.rb
```diff
@@ -265,6 +291,13 @@ def model_with_children(count)
       expect(model_with_children(2).v_total_named_books).to eq(2)
     end
 
+    it "falls back to default for busted virtual association" do
```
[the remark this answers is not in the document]
kbrock: a virtual association that is not written correctly - busted.
e.g. it returns a nil instead of `[]` or `[a, b c]`

lib/active_record/virtual_attributes/virtual_total.rb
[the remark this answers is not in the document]
NickLaMuro: > One thing is for sure, returning different values based upon what what is loaded is a problem

So I think his needs to be addressed then before this is ready to go.

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -122,9 +122,13 @@ def virtual_aggregate_arel(reflection, method_name, column)
...
                         end
-          query       = query.select(arel_column)
-
-          t.grouping(Arel::Nodes::SqlLiteral.new(query.to_sql))
+          query       = t.grouping(Arel::Nodes::SqlLiteral.new(query.select(arel_column).to_sql))
+ 
+          if method_name == :size
+            query
+          else
+            t.grouping(Arel::Nodes::NamedFunction.new('COALESCE', [query, Arel::Nodes::SqlLiteral.new("0")]))
```
kbrock: thinking this should be part of a `default` value clause

spec/virtual_total_spec.rb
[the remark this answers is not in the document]
NickLaMuro: My main question is "busted virtual association" is not very descriptive of what is being referenced, and there is no way for someone who isn't familiar with the code base (or your brain) to figure that out.

lib/active_record/virtual_attributes/virtual_total.rb
NickLaMuro: What do you mean?  As in this snippet:

```
Arel::Nodes::NamedFunction.new('COALESCE', [query, Arel::Nodes::SqlLiteral.new("0")])
```

Should moved to a helper method somewhere?  If so, I think that makes sense, though, arguably could wait for another PR if you don't want to tackle that now.

Up to you.

kbrock:
Something to understand my reasoning behind

old: values.map { |v| v.to_i}.sum 
new: values.compact.sum

I see the second being more similar to the way that sql works.

```sql
create table numbers(num int);
insert into numbers (num) values (4), (8), (null);
select avg(num) from numbers; -- 6
select avg(num) from numbers where num is not null; -- 6
select avg(coalesce(num, 0)) from numbers; -- 4
```

`coalesce(num, 0)` is essentially the same as a sql `to_i` in our case.

method|`f(num)`|`f(num) !null`|`f(num.to_i)`
---|---|---|---
avg|6|6|4
min|4|4|0
sum|10|10|10
max|10|10|10

Note the `sum` case, the only case we use, ends up with the same value

spec/virtual_total_spec.rb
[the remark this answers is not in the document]
kbrock: > falls back to default when virtual association is not written correctly in ruby

lib/active_record/virtual_attributes/virtual_total.rb
kbrock: thinking we need to introduce a `:default => 0` for virtual attributes. if that was present, then we would use the coalesce to ensure the default value

kbrock:
closing in favor of #15

PR #14, opened 2019-04-29
Add tests for virtual totals

kbrock:
need these tests to address more edge cases

if you notice, the same attribute for the same models are returning different values depending upon what is preloaded.

This set of tests document the current behavior, so future code changes can make the behavior changes more obvious (as they work to getting the same values regardless of the loaded values)

### DIFF
--- a/spec/virtual_total_spec.rb
+++ b/spec/virtual_total_spec.rb
@@ -7,7 +7,7 @@ describe VirtualAttributes::VirtualTotal do
 
   describe ".virtual_total" do
     context "with a standard has_many" do
-      it "sorts by total" do
+      it "sorts by total attribute" do
         author2 = Author.create_with_books(2)
         author0 = Author.create_with_books(0)
         author1 = Author.create_with_books(1)
@@ -16,16 +16,34 @@ describe VirtualAttributes::VirtualTotal do
           .to eq([author0, author1, author2].map(&:id))
       end
 
-      it "calculates totals locally" do
-        author0_id = Author.create_with_books(0).id
-        author2_id = Author.create_with_books(2).id
+      it "calculates totals using a query" do
+        author0 = Author.create_with_books(0).reload
+        author2 = Author.create_with_books(2).reload
         expect do
-          expect(Author.find(author0_id).total_books).to eq(0)
-          expect(Author.find(author2_id).total_books).to eq(2)
-        end.to match_query_limit_of(4)
+          expect(author0.total_books).to eq(0)
+          expect(author2.total_books).to eq(2)
+        end.to match_query_limit_of(2)
       end
 
-      it "can bring back totals in primary query" do
+      it "calculates totals with preloaded association" do
+        author_id = Author.create_with_books(2).id
+        author = Author.includes(:books).find(author_id)
+
+        expect do
+          expect(author.total_books).to eq(2)
+        end.to match_query_limit_of(0)
+      end
+
+      it "calculates totals with preloaded associations with no associated records" do
+        author_id = Author.create_with_books(0).id
+        author = Author.includes(:books).find(author_id)
+
+        expect do
+          expect(author.total_books).to eq(0)
+        end.to match_query_limit_of(0)
+      end
+
+      it "calculates totals with attribute" do
         author3 = Author.create_with_books(3)
         author1 = Author.create_with_books(1)
         author2 = Author.create_with_books(2)
@@ -34,6 +52,14 @@ describe VirtualAttributes::VirtualTotal do
           expect(author_query).to match_array([author3, author1, author2])
           expect(author_query.map(&:total_books)).to match_array([3, 1, 2])
         end.to match_query_limit_of(1)
+      end
+
+      it "with no associated records calculates totals with attribute" do
+        Author.create
+        query = Author.select(:id, :total_books).load
+        expect do
+          expect(query.map(&:total_books)).to match_array([0])
+        end.to match_query_limit_of(0)
       end
     end
 
@@ -86,7 +112,7 @@ describe VirtualAttributes::VirtualTotal do
       end
     end
 
-    context "with order clauses in the relation" do
+    context "with a has_many that includes an order" do
       it "sorts by total" do
         skip("fix order in scopes") if ENV["DB"] == "pg"
         author2 = Author.create_with_books(2)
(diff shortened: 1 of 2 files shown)


spec/virtual_total_spec.rb
```diff
@@ -16,16 +16,34 @@
           .to eq([author0, author1, author2].map(&:id))
       end
 
-      it "calculates totals locally" do
-        author0_id = Author.create_with_books(0).id
-        author2_id = Author.create_with_books(2).id
+      it "calculates totals using a query" do
+        author0 = Author.create_with_books(0).reload
+        author2 = Author.create_with_books(2).reload
```
NickLaMuro: Is the `.reload` here just to make sure we are in a fresh state, and then we are only doing the `total_books` query instead of including the `.find` here?

If so, makes sense, just making sure I am understanding the change correctly.

spec/virtual_total_spec.rb
```diff
@@ -245,7 +271,7 @@ class Author < VitualTotalTestBase
       m2 = model_with_children(2)
       mc = m1.class
       expect {
-        ms = mc.select(:id, mc.arel_attribute(:total_books).as("total_books"))
```
NickLaMuro: Huh... why was this here in the first place?

spec/virtual_total_spec.rb
```diff
@@ -328,4 +362,62 @@ def model_with_children(count)
       end
     end
   end
+
+  describe ".virtual_aggregation" do
+    context "with a standard has_many" do
+      let(:authors) { [author, author2, author3]}
```
NickLaMuro: Nitpick:  Space after the `]`

spec/virtual_total_spec.rb
```diff
@@ -35,6 +53,14 @@
           expect(author_query.map(&:total_books)).to match_array([3, 1, 2])
         end.to match_query_limit_of(1)
       end
+
+      it "with no associated records calculates totals with attribute" do
+        Author.create
+        expect do
+          query = Author.select(:id, :total_books).load
+          expect(query.map(&:total_books)).to match_array([0])
+        end.to match_query_limit_of(1)
```
NickLaMuro: Took me a minute to grok this test, but I like that it is added.  👍

spec/virtual_total_spec.rb
```diff
@@ -328,4 +362,62 @@ def model_with_children(count)
...
+
+      it "calculates sum from preloaded association" do
+        author.recently_published_books.load
+        author2.recently_published_books.load
+        author3.recently_published_books.load
+
+        expect do
+          expect(author.sum_recently_published_books_rating).to eq(6)
+          expect(author2.sum_recently_published_books_rating).to eq(5)
+          expect(author3.sum_recently_published_books_rating).to be_nil
```
NickLaMuro: Yeah, this is super important that we have this distinction here when comparing to the one above (results in `0` instead).  Will be important when doing refactoring/fixes.  👍

spec/virtual_total_spec.rb
```diff
@@ -328,4 +362,62 @@ def model_with_children(count)
...
+        end.to match_query_limit_of(0)
+      end
+
+      it "calculates sum from attribute" do
+        authors
+        expect do
+          query = Author.select(:id, :sum_recently_published_books_rating).order(:id).load
+          expect(query.map(&:sum_recently_published_books_rating)).to match_array([6, 5, 0])
+        end.to match_query_limit_of(2)
```
NickLaMuro: What is the second query happening here?  Figured it would be just one.

spec/virtual_total_spec.rb
```diff
@@ -328,4 +362,62 @@ def model_with_children(count)
...
+      end
+
+      it "with no associated records calculates sum from attribute (and preloaded association)" do
+        authors
+        expect do
+          query = Author.includes(:recently_published_books).select(:id, :sum_recently_published_books_rating).load
+          expect(query.map(&:sum_recently_published_books_rating)).to match_array([6, 5, nil])
+        end.to match_query_limit_of(2)
```
NickLaMuro: Same as above.

NickLaMuro: For this one, I could see it being the `.includes`, but I guess that doesn't exist in the one above.

spec/virtual_total_spec.rb
```diff
@@ -265,6 +291,14 @@ def model_with_children(count)
       expect(model_with_children(2).v_total_named_books).to eq(2)
     end
 
+    it "falls back to default when virtual association is not written correctly in ruby" do
+      a = model_with_children(0)
+      # note: this is broken return, but some of our associations return nil instead of [] / none
+      expect(a).to receive(:named_books).and_return(nil)
```
NickLaMuro: Hmm, this (and the note above it) confuses me, but I think it is because I am trying to do this review really quick (and failing).  I assume it is okay, just leaving a note for myself really.

spec/virtual_total_spec.rb
kbrock: I was worried about the creation in `create_with_books` leaving some of the associations pre populated/loaded

spec/virtual_total_spec.rb
kbrock: ikr?

a long time ago (over 2 years?) we were not smart enough to support `select(:virtual_attribute)`

spec/virtual_total_spec.rb
kbrock: @NickLaMuro do you want the load outside of the `expect`?

spec/virtual_total_spec.rb
kbrock: there is a bug in `sum_recently_published_books_rating` that if the value is `nil` then it makes a second query

spec/virtual_total_spec.rb
kbrock: and the answer is the ditto of above

spec/virtual_total_spec.rb
kbrock: we have a `try` in our code that handles a `nil` coming back from an association (virtual I presume).

FWIW/ I don't like it.
struggling keeping this in here.

spec/virtual_total_spec.rb
kbrock: ooh - I extracted. yes this was deceiving

spec/virtual_total_spec.rb
NickLaMuro: This was mostly a "👍 comment" for what it is worth.  Most of my confusion came from where `[0]` was coming from (instead of `[]`), but then I realized it was part of `Author.create` above...

Didn't need to change, but looks like you already did.

spec/virtual_total_spec.rb
NickLaMuro: I like it, and I liked that you droped the `.find` in the `.expect`.  Just making sure I was reading it right.  👍

PR #16, opened 2019-05-01
Support rails 5.1 for preloading values

kbrock:
the way `eager_loading` is handled changes over the rails versions.
It looks like virtual attributes only handles 5.0, and not quite up to snuff with 5.1 (or 5.2)

This brings the eager loading up to par.

This also fixes a bug dealing with includes() passing in array.

/cc @jrafanie this is the issue you were seeing with 5.1 migration

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -30,20 +30,28 @@ module ActiveRecord
           virtual_attribute?(name) || virtual_reflection?(name)
         end
 
-        def remove_virtual_fields(associations)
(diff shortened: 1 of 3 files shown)


kbrock:
looks like polymorphic is broken from the removal of the monkeypatching PR - pulling that back out

kbrock:
I've opted to not fix the code climate (and rubocop) complexity issue. The method was a 14.if I flattened it a little, we could get that back down to a 19, but for now, we're leaving the code as close to the original as possible.

From here, we can focus on removing the monkey patches (which will gain us a lot for stability) and go from there

PR #17, opened 2019-05-01
tests for null checks, requires and includes, and query expectations

kbrock:
- tests are more consistent for the null checks.
- tests around requires and includes
- replaced `not_raise_exception` with query expectations. so rather than just say "works", it now also ensures the correct data was fetched.
- `create_books` actually returns books to make a few specs easier to write
- fixed 'named_books' `:uses` clause. now `includes(:named_books)` actually does something (this was detected from the `not_raise_exception` change)

### DIFF
--- a/spec/db/models.rb
+++ b/spec/db/models.rb
@@ -33,7 +33,7 @@ class Author < VitualTotalTestBase
     books.select { |b| b.name }
   end
 
-  virtual_has_many :named_books, :class_name => "Book"
+  virtual_has_many :named_books, :class_name => "Book", :uses => :books
   virtual_total :total_named_books, :named_books
   alias v_total_named_books total_named_books
 
--- a/spec/db/models.rb
+++ b/spec/db/models.rb
@@ -45,17 +45,13 @@ class Author < VitualTotalTestBase
     t.grouping(Arel::Nodes::NamedFunction.new('COALESCE', [t[:nickname], t[:name]]))
   end
 
-  def self.create_with_books(count = 0)
+  def self.create_with_books(count)
     create!(:name => "foo").tap { |author| author.create_books(count) }
   end
 
   def create_books(count, create_attrs = {})
-    count.times do
-      attrs = {
-        :name   => "bar",
-        :author => self,
-      }.merge(create_attrs)
-      Book.create(attrs)
+    Array.new(count) do
+      books.create({:name => "bar"}.merge(create_attrs))
     end
   end
 end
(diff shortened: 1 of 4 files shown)


spec/db/models.rb
```diff
@@ -33,7 +33,7 @@ def named_books
     books.select { |b| b.name }
   end
 
-  virtual_has_many :named_books, :class_name => "Book"
+  virtual_has_many :named_books, :class_name => "Book", :uses => :books
```
kbrock: now `includes(:named_books)` actually does something

spec/db/models.rb
```diff
@@ -45,17 +45,13 @@ def nick_or_name
...
   end
 
   def create_books(count, create_attrs = {})
-    count.times do
-      attrs = {
-        :name   => "bar",
-        :author => self,
-      }.merge(create_attrs)
-      Book.create(attrs)
+    Array.new(count) do
```
kbrock: improved for specs (`create_books` should return books...)

spec/support/match_query_limit_of.rb
```diff
@@ -4,7 +4,7 @@
 #   expect { MyModel.do_the_queries }.to match_query_limit_of(5)
 
 RSpec::Matchers.define :match_query_limit_of do |expected|
-  match do |block|
+  match(:notify_expectation_failures => true) do |block|
```
kbrock: makes specs easier since now, embedded exceptions are raised (vs saying it failed the 0 query count - which was useless)

spec/virtual_attributes_spec.rb
```diff
@@ -933,47 +933,79 @@ def col2
...
-        expect { Author.includes(:nick_or_name => {}, :bookmarks => :book).load }.not_to raise_error
+        expect { Author.includes(:nick_or_name => {}).load }.to match_query_limit_of(1)
+        expect { Author.includes(:nick_or_name => {}, :bookmarks => :book).load }.to match_query_limit_of(4)
       end
     end
 
     context "virtual reflection" do
       it "as Symbol" do
-        expect { Author.includes(:named_books).load }.not_to raise_error
+        expect { Author.includes(:named_books).load }.to match_query_limit_of(2)
```
kbrock: this actually caught the missing `:uses` clause

jrafanie:
LGTM, do we want to merge this or see if any proposals on the fix would require changes here?  In other words, are these good to go?

kbrock:
Yea, want tests to define the current behavior - we will change then when we want different behavior.

Especially with the 0 return values and the references.
They are definitely highlighting issues here.

PR #18, opened 2019-05-01
code climate - remove duplicate matcher

kbrock:
alias a matcher instead of duplicating the code


kbrock:
fixed up the missing parens

PR #20, opened 2019-05-07
[WIP] require types defined

kbrock:
running into deadlocks with this.

The referenced schema (and class) need to be loaded for this class to be fully defined.

Unfortunately, that then requires that the other foreign classes also be loaded.

This is causing race conditions with the locks around loading classes.

One option is to punt on the type discovery. Which is really a shame, but it remove the deadlocks

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_delegates.rb
+++ b/lib/active_record/virtual_attributes/virtual_delegates.rb
@@ -143,9 +143,14 @@ module ActiveRecord
         #   See select_from_alias for examples
 
         def virtual_delegate_arel(col, to_ref)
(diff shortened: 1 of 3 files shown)


jrafanie:
@d-m-u please review

kbrock:
Having second thoughts on this one

d-m-u:
regardless of whether you end up going with something else, here's the bz: 
https://bugzilla.redhat.com/show_bug.cgi?id=1700378

kbrock:
rebased this on top of 21

The difference for this being that the `:type` is required to be specified

kbrock:
I prefer backwards compatible. I can resurrect if need be. Deferring to #21

PR #22, opened 2019-05-10
add tests for includes().references() focusing on data types

kbrock:
### Overview

`includes().references()` is failing on rails 5.1

While working through the code in #16 , I wanted to put more tests around the way the uses and the references or includes interact.

### This PR

- This pulls the preload / virtual_includes logic out into separate tests.
- It also introduces many many more combinations, focusing on the data types more
- It introduces a matcher focused on whether an attribute is preloaded.
- there are a few tests commented out (they fail due to bugs) - will resolve in a future PR
- changed one PR to bring back data, this started causing it to fail on 5.1 (the no data was masking the failure)


kbrock:
cop fixed

kbrock:
ugh `|| {}` is hard :( sorry for the bonus pushes

spec/support/preload_values.rb
```diff
@@ -0,0 +1,33 @@
+# Example usage:
+#   expect { MyModel.do_the_queries }.to preload_values(:attribute, value)
```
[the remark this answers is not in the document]
NickLaMuro: Also, `value` in this definition should probably be `value[s]` as it seems based on the implementation that it can either be a single value, or array of values.

spec/virtual_includes_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+  before do
+    Author.destroy_all
+    Book.destroy_all
+    Author.create_with_books(3).books.first.create_bookmarks(2)
+  end
+
+  # books.map(&:author_name)
+  let(:foos) { %w[foo foo foo] }
+  let(:named_books) { Book.where.not(:name => nil).order(:id).load }
+  # # authors.map(&:named_books) (each of 1 author has an array of named books)
```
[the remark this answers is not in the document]
NickLaMuro: Also, looking at the implementation of the `preload_values` matcher, specifically:

```ruby
expected_values = Array(expected_values) unless expected_values.kind_of?(Array)
```

Is `named_bookss` even needed, since it would just get mapped to an array anyway?

spec/support/preload_values.rb
```diff
@@ -0,0 +1,33 @@
...
+
+RSpec::Matchers.define :preload_values do |field, expected_values|
+  match(:notify_expectation_failures => true) do |block|
+    results = block.call.order(:id).load
+
+    @count = ActiveRecord::QueryCounter.count do
+      if results.respond_to?(:map)
+        expected_values = Array(expected_values) unless expected_values.kind_of?(Array)
```
NickLaMuro: I mentioned about this line elsewhere, but if we are already checking against `:map`, doesn't that assume the use of an `Array`?

spec/virtual_includes_spec.rb
```diff
@@ -0,0 +1,180 @@
+describe ActiveRecord::VirtualAttributes::VirtualIncludes do
+  before do
+    Author.destroy_all
+    Book.destroy_all
+    Author.create_with_books(3).books.first.create_bookmarks(2)
+  end
+
+  # books.map(&:author_name)
+  let(:foos) { %w[foo foo foo] }
```
NickLaMuro: Can we call this `author_names` or something?  `foos` isn't very descriptive, and I feel like I would always have to look this up when viewing this in the future.

NickLaMuro: This might need to be another commit, but you could also consider doing in `Author`:

```diff
- def self.create_with_books(count, name=foo)
-   create!(:name => "foo").tap { |author| author.create_books(count) }
+ def self.create_with_books(count, name=foo)
+   create!(:name => name).tap { |author| author.create_books(count) }
```

Which would allow you to change the about line to something like:

```ruby
Author.create_with_books(3, "Keenan").books.first.create_bookmarks(2)
```

And the `let(:foos)` to:

```ruby
let(:author_names) { %w[Keenan Keenan Keenan] }
```

Though, might want to choose a different name if you don't want to look too narcissistic...

spec/support/preload_values.rb
```diff
@@ -0,0 +1,33 @@
...
+        expect(actual).to eq(expected_values)
+      else
+        expect(results.send(field)).to eq(expected_values)
+      end
+    end
+    @count == 0
+  end
+
+  failure_message do |_actual|
+    "Expected to preload values (#{@count} queries)"
```
NickLaMuro: I feel like this might look weird, because you are testing two things:

- that the values are what we expect
- the query count is `0` (after the `.load` is called above)

Can you provide some output for failure cases for each of the above?  Do they seem readable?

spec/virtual_attributes_spec.rb
```diff
@@ -971,91 +971,6 @@ def col2
...
   end
 
-  context "preloading" do
-    before do
-      Author.destroy_all
-      Book.destroy_all
-      Author.create_with_books(3).books.first.create_bookmarks(2)
-    end
-
-    context "virtual column" do
```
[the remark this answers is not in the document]
NickLaMuro: Okay, after looking at a8484ca , I think I have a better understanding how you refactored things.

That said, on their own, they were still a bit confusing to parse (at least for my little head).

spec/virtual_includes_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      expect { Author.includes([:named_books, :bookmarks]).load }.to match_query_limit_of(3)
+    end
+
+    it "as Hash" do
+      expect { Author.includes(:named_books => {}).load }.to match_query_limit_of(2)
+      expect { Author.includes(:named_books => {}, :bookmarks => :book).load }.to match_query_limit_of(4)
+    end
+  end
+
+  # FIX
```
NickLaMuro: FYI:  There any context with this "FIX" comment?  It just disappears 2 commits later.

What did you want to fix?  Was it fixed?

spec/support/preload_values.rb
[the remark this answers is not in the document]
kbrock: Hope docs did all this.
the definition did change a little as your question felt a little like a "by the way, did you also check..."

spec/virtual_includes_spec.rb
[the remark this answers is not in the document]
kbrock: it didn't kick in. It couldn't distinguish between `[value]` and `[[value]]`

thanks for info on this one.

spec/support/preload_values.rb
kbrock: For some tests, I found myself wanting to pass in a single value in some cases instead of an array.

I've fixed this up to work better with passing in a single value.

So the code has changed, but yes, the logic is basically the same.
(and yes, I was still not able to get the plural version working)

spec/support/preload_values.rb
kbrock: since we use `notify_expectation_failures` the inner expect will report it's own failure.
So this failure is actually just when the record count does not work.

spec/virtual_attributes_spec.rb
```diff
@@ -971,91 +971,6 @@ def col2
...
-      end
-
-      it "as Hash" do
-        expect { Author.includes(:named_books => {}).load }.to match_query_limit_of(2)
-        expect { Author.includes(:named_books => {}, :bookmarks => :book).load }.to match_query_limit_of(4)
-      end
-    end
-
-    it "nested virtual fields" do
-      expect { Author.includes(:books => :author_name).load }.to match_query_limit_of(3)
```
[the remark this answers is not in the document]
kbrock: yea, this was a duplicate spec. I'll make sure the comments show that

spec/virtual_includes_spec.rb
kbrock: thnx. I added context around it.

it had an miq column in there that wasn't valid

spec/support/preload_values.rb
kbrock: > "Expected to preload #{@field} but executed #{@count} queries instead"

I'll look at these failure cases

spec/virtual_includes_spec.rb
kbrock: I fixed `author_name` - it reads much better.
With the change to the way the array works, it is much more straightforward.

I avoided making a real factory for the specs. they should not live in the model but unlike traditional projects, the models are 100% for specs.

spec/db/models.rb
```diff
@@ -27,24 +27,45 @@ class Author < VitualTotalTestBase
   virtual_total :total_recently_published_books, :recently_published_books
   virtual_aggregate :sum_recently_published_books_rating, :recently_published_books, :sum, :rating
 
-  # virtual_total using a virtual_has_many
+  # virtual_total using a virtual_has_many with an array of associations
+  # LAZY: sorry for lack of creativity, this could have been done with a where
```
NickLaMuro: I wouldn't call this "LAZY", but I would say it is a use case of a poorly optimized `virtual_attribute`.

This comment is also redundant (regarding "creativity") when paired with the one inside the method.

* * *

Idea in the future to replace this would be to do something like `book_type`, where "valid types" are an array in ruby, and they are formatted via `formatted_book_type`, and "Unknown" types that aren't in the array are just formatted to "Other".

But I think that idea deserves another PR.

spec/db/models.rb
```diff
@@ -27,24 +27,45 @@ class Author < VitualTotalTestBase
   virtual_total :total_recently_published_books, :recently_published_books
   virtual_aggregate :sum_recently_published_books_rating, :recently_published_books, :sum, :rating
 
-  # virtual_total using a virtual_has_many
+  # virtual_total using a virtual_has_many with an array of associations
```
NickLaMuro: Is this really a `virtual_total`?  Did I add this comment?  If so... sorry...

NickLaMuro: Looks like I didn't add this:

https://github.com/ManageIQ/activerecord-virtual_attributes/commit/9308002c7dd9c3bd4e42c32c552f0e63feedf8bc#diff-4520e22fc7e8d7545816ca398c24a778R55

But also, it isn't a `virtual_total`...

https://github.com/ManageIQ/activerecord-virtual_attributes/commit/9308002c7dd9c3bd4e42c32c552f0e63feedf8bc#diff-4520e22fc7e8d7545816ca398c24a778R61

So yeah...

spec/db/models.rb
```diff
@@ -27,24 +27,45 @@ class Author < VitualTotalTestBase
...
+  # LAZY: sorry for lack of creativity, this could have been done with a where
   def named_books
     # I didn't have the creativity needed to find a good ruby only check here
     books.select { |b| b.name }
   end
 
+  # attribute with uses that depends upon a virtual column in another model.
+  # this tests a uses with a virtual column, a hash uses.
+  # LAZY: sorry for lack of creativity, this could be done with pure sql, or a delegate, but aiming for those usecases
```
NickLaMuro: Here too:  I would just comment that this is a "poorly optimized use case" that we support.

spec/virtual_includes_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+      skip("AR 5.1 not including properly") if ActiveRecord.version.to_s >= "5.1"
+      expect { Author.includes(:first_book_author_name).references(:first_book_author_name => {}) }.to preload_values(:first_book_author_name, author_name)
+      expect { Author.includes([:first_book_author_name]).references(:first_book_author_name => {}) }.to preload_values(:first_book_author_name, author_name)
+      expect { Author.includes(:first_book_author_name => {}).references(:first_book_author_name => {}) }.to preload_values(:first_book_author_name, author_name)
+    end
+
+    it "uses included associations" do
+      skip("AR 5.1 not including properly") if ActiveRecord.version.to_s >= "5.1"
+      expect { Author.includes(:books => :author).references(:books => {:author => {}}) }.to preload_values(:first_book_author_name, author_name)
+      #expect { Author.includes(:books => [:author]).references(:books => {:author => {}}) }.to preload_values(:first_book_author_name, author_name)
```
NickLaMuro: I do agree with `rubocop` here...

spec/support/preload_values.rb
```diff
@@ -0,0 +1,51 @@
...
+# This takes a relation or array as a block or argument.
+# It runs order by id if it is not already ordered
+# It loads the relation if necessary (by running load)
+#
+# It then ensures the correct values, and in the process counts queries to ensure they were preloaded
+#
+# If a single value is passed it, it assumes all records have the same value
+#
+# Example usage:
+#   expect(MyModel.includes(:relation)).to preload_values(:attribute, values)
```
NickLaMuro: A few things like this:

I think it would help having two examples:  One with an array and one with a single value:

```ruby
expect(MyModel.includes(:relation)).to preload_values(:attribute, "foo")
expect(MyModel.includes(:relation)).to preload_values(:attribute, %w[foo bar baz])
```

Second, your example uses the "non-block" form, but you never actually use it in the form in the specs.  My question here is actually:  does this make sense to support both forms?  Seems like it might be simpler to pick one and go with that.  Personally I think the non-block is easier to type out, but I don't think it matters one way or the other.

spec/virtual_attributes_spec.rb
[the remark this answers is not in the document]
kbrock: Hope the extra commits have added more context here.

spec/virtual_includes_spec.rb
```diff
@@ -14,14 +14,22 @@
     it "preloads virtual_attribute (:uses => nil) (with a NO OP)" do
       expect { Author.includes(:nick_or_name) }.to preload_values(:nick_or_name, author_name)
       expect { Author.includes([:nick_or_name]) }.to preload_values(:nick_or_name, author_name)
-
```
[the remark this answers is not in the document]
kbrock: aah - no I had to put it in here.

when I deleted this line, the diffs were all over the place and totally confusing. I kinda had to do it this way :(
(I did this trick twice)

spec/virtual_includes_spec.rb
```diff
@@ -39,6 +39,20 @@
 
   end
 
+  # references follow a different path than just includes
+  context "preload virtual_attribute with includes.references" do
+    it "preload virtual_attribute (delegate defines :uses => :author)" do
+      skip("AR 5.1 not including properly") if ActiveRecord.version.to_s >= "5.1"
+      expect { Book.includes(:author_name).references(:author_name => {}) }.to preload_values(:author_name, author_name)
```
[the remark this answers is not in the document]
kbrock: the non hash version blows up.
btw/ we only use the hash version in our code

spec/virtual_includes_spec.rb
```diff
@@ -14,14 +14,22 @@
     it "preloads virtual_attribute (:uses => nil) (with a NO OP)" do
       expect { Author.includes(:nick_or_name) }.to preload_values(:nick_or_name, author_name)
       expect { Author.includes([:nick_or_name]) }.to preload_values(:nick_or_name, author_name)
-
       expect { Author.includes(:nick_or_name => {}) }.to preload_values(:nick_or_name, author_name)
     end
 
+    it "preload virtual_attribute (delegate defines :uses => :author)" do
```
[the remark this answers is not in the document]
kbrock: ugh - this is a lot of commits. will do this and hope not to botch too much

spec/support/preload_values.rb
kbrock: Yea, at first, the code did more. it counted queries before and stuff like that.

Now, the paren version works just as well and is easier to read.
may just circle back and change this in bulk rather than try and fix the slew of commits (with cascading git merge failures)

spec/db/models.rb
kbrock: I made the comment more clear. this is here to be referenced by a virtual total

spec/virtual_includes_spec.rb
```diff
@@ -39,6 +39,13 @@
 
   end
 
+
+  it "preload virtual_attribute in :include when :conditions are also present in calculations" do
+    skip("AR 5.1 not including properly") if ActiveRecord.version.to_s >= "5.1"
+    expect { Book.includes([:author_name, :author]).references(:author).where("authors.name = 'foo'") }.to preload_values(:author_name, author_name)
```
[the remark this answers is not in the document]
kbrock: I fixed to use `#{author_name}` - hope this is more obvious.

spec/virtual_includes_spec.rb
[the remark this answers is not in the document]
kbrock: fun, only references(array) or references(symbol) is supported by rails - yet we tend to always use the hash syntax in Rbac

spec/db/models.rb
```diff
@@ -33,7 +33,13 @@ def named_books
     books.select { |b| b.name }
   end
 
+  # and again, no creativity - and obviously it has an author's name...
+  def books_with_authors
```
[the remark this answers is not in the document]
kbrock: example: https://github.com/kbrock/manageiq/blob/master/app/models/disk.rb#L9

not sure if that is valid, but we have a number of `:uses => virtual_attribute`

spec/support/preload_values.rb
NickLaMuro: Okay, yeah, I looked at this yesterday and it seemed fine.  What you have is good.

kbrock:
kicking - hoping to fix rubocop configuration file complaint

spec/virtual_includes_spec.rb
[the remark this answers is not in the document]
NickLaMuro: It does, actually!

The commit message is still feels a little cryptic, but at this point leave it as is.  This is also why I am just mentioning it and not giving you and example so you don't go and change it.  😉

PR #23, opened 2019-05-13
Support rails 5.1 for preloading values. Also reduce monkey patching

kbrock:
### Description

Alternate to #16  (it builds upon it really)

Instead of monkey patching internal methods, this patches the methods where the developer passes in the virtual Attributes.

### Overview

We call `includes().references().select().order()` with virtual_attributes, or virtual_includes.

Some code needs to change the virtual attributes we pass in, over to the real models that rails can actually use.

### Before

We store the passed in values in `includes()` and `references()`. At query time we convert those values to the appropriate values.

Unfortunately, these internal classes change a bunch over the rails versions (and sometimes within a rails bug version).

### After

We change the passed in values in `includes()` and `references()` to be the actual values we want to use at query time. At query time there is no need to do anything special.

These methods are relatively public in nature and do not change that often. This makes maintaining multiple versions much easier.

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -71,26 +71,6 @@ module ActiveRecord
   end
 
   module Associations
-    class Preloader
-      prepend(Module.new {
-        def preloaders_for_one(association, records, scope)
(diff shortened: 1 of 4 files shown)


kbrock:
putting this commit into #16

PR #24, opened 2019-05-15
Fix some trivial cops

kbrock:
fix some trivial cops

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -1,0 +1,8 @@
+---
+extends: relaxed
+
+rules:
+  indentation:
+    indent-sequences: false
+  line-length:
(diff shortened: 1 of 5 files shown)


kbrock:
Kicking

jrafanie:
Let's see if this greens codeclimate on master after merge.  It looks ok.

.yamllint
```diff
@@ -0,0 +1,14 @@
+---
+ignore: |
+  /lib/generators/**/templates/**
```
Fryguy: Minor, but you don't have generators here, not spec/tools/openstack_data, so those lines can be removed.

PR #25, opened 2019-05-17
Fix poly references

kbrock:
Fixes an issue when an includes references a column without using hashes.

invalid columns were silently failing in many cases.
Which is a problem because it was missing some includes.
This plugs that hole so we can properly fix these problems when we see them

```ruby
# before:
Author.includes(:books => [:author]).references(:books => {:author => {}}).map(&:name)

     ActiveRecord::ConfigurationError:
       nil
     # /Users/kbrock/.gem/ruby/2.5.5/gems/activerecord-5.0.7.2/lib/active_record/associations/join_dependency.rb:70:in `walk_tree'
     # /Users/kbrock/.gem/ruby/2.5.5/gems/activerecord-5.0.7.2/lib/active_record/associations/join_dependency.rb:67:in `block in walk_tree'
     # /Users/kbrock/.gem/ruby/2.5.5/gems/activerecord-5.0.7.2/lib/active_record/associations/join_dependency.rb:65:in `each'
     # /Users/kbrock/.gem/ruby/2.5.5/gems/activerecord-5.0.7.2/lib/active_record/associations/join_dependency.rb:65:in `walk_tree'
...

# after:

Author.includes(:books => [:author]).references(:books => {:author => {}}).to_a
["foo", "foo", "foo"]
```

Also

```ruby
# before:
 Author.includes(:books => {:invalid => {}}).references(:books).map(&:name)
["foo", "foo", "foo"]

# after:
 Author.includes(:books => {:invalid => {}}).references(:books).map(&:name)
     ActiveRecord::ConfigurationError:
       Can't join 'Author' to association named 'invalid'; perhaps you misspelled it?
     # /Users/kbrock/.gem/ruby/2.5.5/gems/activerecord-5.0.7.2/lib/active_record/associations/join_dependency.rb:231:in `find_reflection'
     # /Users/kbrock/.gem/ruby/2.5.5/gems/activerecord-5.0.7.2/lib/active_record/associations/join_dependency.rb:236:in `block in build'
```

### DIFF
--- a/spec/virtual_includes_spec.rb
+++ b/spec/virtual_includes_spec.rb
@@ -53,6 +53,11 @@ describe ActiveRecord::VirtualAttributes::VirtualIncludes do
       expect(Author.includes(:first_book_author_name => {})).to preload_values(:first_book_author_name, author_name)
     end
 
+    it "preloads through polymorphic" do
+      books = Book.includes(:author_or_bookmark => :total_books).load
+      expect { expect(books.map { |b| b.author_or_bookmark.total_books }).to eq([3, 3, 3]) }.to match_query_limit_of(0)
+    end
+
     it "uses included associations" do
       expect(Author.includes(:books => :author)).to preload_values(:first_book_author_name, author_name)
       expect(Author.includes(:books => {:author => {}})).to preload_values(:first_book_author_name, author_name)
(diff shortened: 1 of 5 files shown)


spec/virtual_includes_spec.rb
```diff
@@ -101,28 +101,35 @@
...
+      expect(Author.includes(:books => :author_name).references(:books => {:author_name => {}})).to preload_values(:first_book_author_name, author_name)
+      expect(Author.includes(:books => [:author_name]).references(:books => {:author_name => {}})).to preload_values(:first_book_author_name, author_name)
       expect(Author.includes(:books => {:author_name => {}}).references(:books => {:author_name => {}})).to preload_values(:first_book_author_name, author_name)
-      # inc = Author.virtual_includes(:first_book_author_name)
-      # expect(Author.includes(inc).references(inc)).to preload_values(:first_book_author_name, author_name)
+      inc = Author.virtual_includes(:first_book_author_name)
+      expect(Author.includes(inc).references(inc)).to preload_values(:first_book_author_name, author_name)
+    end
+
+    it "detects errors" do
```
jrafanie: Can you add one example to this group where it's correct? and change this to "raises if includes don't have a proper uses clause?

jrafanie: I think a good example in contrast to the failure ones will better demonstrate what we're doing here.

kbrock: something like this?

```ruby
      expect { Author.includes(:books).references(:books).load}.not_to raise_error
```

jrafanie: Right

kbrock:
fixed cops

kbrock:
added the polymorphic tests here. they were fixed by the bugfix as well.

PR #26, opened 2019-05-17
Test virtual using virtual attributes

kbrock:
Adding specs around a virtual attribute that uses another virtual attribute

this ensures that we recursively resolve the uses clauses


kbrock:
moved polymorphic test over to 25 - since that was messing with the hash returned for polymorphic (and fixed a bug around it)

added rails 5.1 issues here and re-pushed. should be good to go

PR #27, opened 2019-05-31
Column aliases

kbrock:
ManageIQ/manageiq: Custom Attributes define :arel with an alias.

This allows the custom attributes to be defined without an alias
And this fixes an edge case when an alias is defined

To be honest, we should probably remove the custom alias for the custom attributes when defining the virtual attributes - but this PR is needed for that case as well.

Also, this refactored that method to make rubocop happy

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -200,6 +200,7 @@ module ActiveRecord
       end
 
       # From ActiveRecord::QueryMethods
+      # would be nice to get this monkey patch into arel_column
       def select(*fields)
         return super if block_given? || fields.empty?
         # support virtual attributes by adding an alias to the sql phrase for the column
@@ -207,14 +208,12 @@ module ActiveRecord
         # this code is based upon _select()
         fields.flatten!
         fields.map! do |field|
-          if virtual_attribute?(field) && (arel = klass.arel_attribute(field))
-            if arel.respond_to?(:as) && !arel.try(:alias)
-              arel.as(connection.quote_column_name(field.to_s))
-            else
-              arel
-            end
-          else
+          if !virtual_attribute?(field) || !(arel = klass.arel_attribute(field))
             field
+          elsif arel.respond_to?(:as) && !arel.try(:alias)
+            arel.as(connection.quote_column_name(field.to_s))
+          else
+            arel
           end
         end
         # end support virtual attributes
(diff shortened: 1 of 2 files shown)


kbrock:
re code climate: that method has not been touched. ignoring

jrafanie:
@NickLaMuro can you 👀 the second commit... I'm not sure ...

@kbrock what are we giving up by it not working for `order` and `where` but will work with `select`? Does this break things, regress?  I guess I'm  not sure if this breaks things or just doesn't support aliases completely.

kbrock:
> @kbrock what are we giving up by it not working for order and where but will work with select? Does this break things, regress? I guess I'm not sure if this breaks things or just doesn't support aliases completely.

@jrafanie Good question. We really should not be using an `alias` there. You are right. But we do use that in miq's custom attributes. So I'm putting in this change to fix custom attributes, and that will allow us to keep miq as it is, or to improve it.

NickLaMuro:
## Nick's 2cents™

High level, I gave this response Keenan in a DM about this a few days ago:

> the point wasn't you shouldn't fix those issues, but instead that possibly you shouldn't fix those #rightNow™

And also linked to this [image](http://gifs.chrisarcand.com/changing-a-lightbulb.gif)...

* * *

But if I recall correctly, this was specifically being targeted in response to issue when trying to enable `inline_view?` for all reports in our project:

https://github.com/kbrock/manageiq/blob/0e5652b26577d65dca2579543a2adee7b144798b/lib/rbac/filterer.rb#L276

Which I think is something easier solved by fixing for _**US**_ (MIQ) first, and deciding what to do here the #rightWay™ when it is understood better.  Specifically:

> "I'm not a big fan of this approach as it will break use of this column for where and order clauses, but will still work in select clauses" (from second commit message)

Can this then not be fixed here and instead in MIQ for our specific use case? (see link above).  If you yourself don't like this approach, I am having a hard time even starting to review this further.

I think the first commit, however, is completely fine and I think makes sense to be merged in on it's own since I don't see it hurting anything.

kbrock:
>>"I'm not a big fan of this approach as it will break use of this column for where and order clauses, but will still work in select clauses" (from second commit message)
>> -- @kbrock

>Can this then not be fixed here and instead in MIQ for our specific use case? (see link above). If you yourself don't like this approach, I am having a hard time even starting to review this further.
> --  @NickLaMuro 

Yes, I'm not a big fan of how we coded the alias in MIQ in the first place.
The second commit allows us to leave the code in MIQ that I don't like.

If we just do the first commit, then we can change MIQ to remove the alias from the custom attributes and this will all work.

kbrock:
I want to keep this PR for prosperity case, oped #28 with just the first commit.

Thanks

PR #28, opened 2019-06-02
extract replace_virtual_field_hash

kbrock:
make code climate happy

not impressed with this extraction but...


jrafanie:
> make code climate happy
> 
> not impressed with this extraction but...

You're very convincing...  🤣

PR #30, opened 2019-06-05
quote column aliases

kbrock:
properly quote virtual attribute names

Sometimes the attribute name is not sql friendly.
the will properly quote the attribute name

In our case, custom attributes sometimes have non sql friendly name for the virtual attribute. this escapes the code and makes us all happy

NOTE: this is the first commit from #27 -- I wanted that code to stick around so I created this smaller PR


kbrock:
>But if I recall correctly, this was specifically being targeted in response to issue when trying to enable inline_view? for all reports in our project:
>
>https://github.com/kbrock/manageiq/blob/0e5652b26577d65dca2579543a2adee7b144798b/lib/rbac/filterer.rb#L276
>
> [...]
>
>I think the first commit, however, is completely fine and I think makes sense to be merged in on it's own since I don't see it hurting anything.
>
>--@NickLaMuro

PR #31, opened 2019-06-05
refactor: simplify virtual_aggregate and virtual_total

kbrock:
This fixes Code Climate issues:

```
Method define_virtual_aggregate_method has a Cognitive Complexity of 20
(exceeds 11 allowed). Consider refactoring. 
```

### what was done

- This makes no functional changes.
- It splits a method into 3 methods - to appease CodeClimate and me.
- added virtual_aggregate :size method to handle cases where someone used `virtual_aggregates` instead of `virtual_totals` (not sure if we should support this, but put in just in case)

### reviewing tips

viewing with `ignoring whitespace` makes it a little more obvious that stuff is just moved around

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_total.rb
+++ b/lib/active_record/virtual_attributes/virtual_total.rb
@@ -25,7 +25,7 @@ module VirtualAttributes
       #   # arel == (SELECT COUNT(*) FROM vms where ems.id = vms.ems_id)
       #
       def virtual_total(name, relation, options = {})
-        define_virtual_aggregate_method(name, relation, :size, nil)
+        define_virtual_size_method(name, relation)
         define_virtual_aggregate_attribute(name, relation, :size, nil, options)
       end
 
@@ -75,23 +75,23 @@ module VirtualAttributes
         end
       end
 
+      def define_virtual_size_method(name, relation)
+        define_method(name) do
+          (attribute_present?(name) ? self[name] : nil) || send(relation).try(:size) || 0
+        end
+      end
+
       def define_virtual_aggregate_method(name, relation, method_name, column)
-        if method_name == :size
-          define_method(name) do
-            (attribute_present?(name) ? self[name] : nil) || send(relation).try(:size) || 0
-          end
-        else
-          define_method(name) do
-            (attribute_present?(name) ? self[name] : nil) ||
-              begin
-                rel = send(relation)
-                if rel.loaded?
-                  rel.blank? ? nil : (rel.map { |t| t.send(column).to_i } || 0).send(method_name)
-                else
-                  rel.try(method_name, column) || 0
-                end
+        define_method(name) do
+          (attribute_present?(name) ? self[name] : nil) ||
+            begin
+              rel = send(relation)
+              if rel.loaded?
+                rel.blank? ? nil : (rel.map { |t| t.send(column).to_i } || 0).send(method_name)
+              else
+                rel.try(method_name, column) || 0
               end
-          end
+            end
         end
       end
 
--- a/lib/active_record/virtual_attributes/virtual_total.rb
+++ b/lib/active_record/virtual_attributes/virtual_total.rb
@@ -53,6 +53,8 @@ module VirtualAttributes
       #    # arel => (SELECT sum("disks"."size") where "hardware"."id" = "disks"."hardware_id")
 
       def virtual_aggregate(name, relation, method_name = :sum, column = nil, options = {})
+        return define_virtual_total(name, relation, options) if method_name == :size
+
         define_virtual_aggregate_method(name, relation, method_name, column)
         define_virtual_aggregate_attribute(name, relation, method_name, column, options)
       end
--- a/lib/active_record/virtual_attributes/virtual_total.rb
+++ b/lib/active_record/virtual_attributes/virtual_total.rb
@@ -25,7 +25,8 @@ module VirtualAttributes
       #   # arel == (SELECT COUNT(*) FROM vms where ems.id = vms.ems_id)
       #
       def virtual_total(name, relation, options = {})
-        virtual_aggregate(name, relation, :size, nil, options)
+        define_virtual_aggregate_method(name, relation, :size, nil)
+        define_virtual_aggregate_attribute(name, relation, :size, nil, options)
       end
 
       # define an attribute to calculate the sum of a has may relationship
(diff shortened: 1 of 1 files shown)


kbrock:
We have the following virtual aggregates in MIQ. All fields are integers.

```
class Ems
  virtual_aggregate :total_vcpus,            :hosts, :sum, :total_vcpus
  virtual_aggregate :total_memory,           :hosts, :sum, :ram_size
  virtual_aggregate :total_cloud_vcpus,      :vms,   :sum, :cpu_total_cores
  virtual_aggregate :total_cloud_memory,     :vms,   :sum, :ram_size
end

class Hardware
  virtual_aggregate :used_disk_storage,      :disks, :sum, :used_disk_storage
  virtual_aggregate :allocated_disk_storage, :disks, :sum, :size
end
```

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -69,24 +75,25 @@ def virtual_aggregate(name, relation, method_name = :sum, column = nil, options
...
-                  rel.try(method_name, arel_column) || 0
-                end
-              end
+        define_method(name) do
+          if attribute_present?(name)
+            self[name]
+          else
+            rel = send(relation)
+            if rel.loaded?
+              rel.blank? ? nil : rel.map { |t| t.send(column) }.compact.send(method_name)
```
kbrock: this is the only line that changed in this method. I mentioned it in the opening comment.

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -69,24 +75,25 @@ def virtual_aggregate(name, relation, method_name = :sum, column = nil, options
         end
       end
 
+      def define_virtual_size_method(name, relation)
+        define_method(name) do
+          (has_attribute?(name) ? self[name] : send(relation).try(:size)) || 0
```
kbrock: this changed from `attribute_present` to `has_attribute`
Since a sql `count(*)` is never `null`, this will never be `nil`. So the `nil` check is pointless.

lib/active_record/virtual_attributes/virtual_total.rb
kbrock: I removed this refactor

lib/active_record/virtual_attributes/virtual_total.rb
kbrock: This comment is covered in commit `virtual_total nil attribute does not exec query`

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -75,25 +75,25 @@ def define_virtual_aggregate_attribute(name, relation, method_name, column, opti
         end
       end
 
+      def define_virtual_size_method(name, relation)
+        define_method(name) do
+          (attribute_present?(name) ? self[name] : nil) || send(relation).try(:size) || 0
+        end
+      end
+
       def define_virtual_aggregate_method(name, relation, method_name, column)
```
NickLaMuro: Could this technically be a problem if someone passes `:size` to `virtual_aggregate`?  I am not against the change, but it does seem to make it a bit if a breaking change since there is a functional difference as a result.

kbrock: ok.

we own all the callers to this functionality.

55 calls are to `virtual_total`
6 calls are to `virtual_aggregate`, all of which are for `:sum`.

Do we want to ignore this case, or add an escape clause?

```ruby
  def define_virtual_aggregate_method(name, relation, method_name, column)
    # escape clause
    return define_virtual_total(name, relation, options) if method_name == :size

    define_virtual_aggregate_method(name, relation, method_name, column)
    define_virtual_aggregate_attribute(name, relation, method_name, column, options)
 end
```

kbrock: @NickLaMuro I'm leaning towards just ignoring that case / aka let the case that we never call break.

NickLaMuro: Preface:  I don't have merge rights on this project or really have much of a say in it's direction, so this is mostly an opinion response.

(also, sorry I missed your comment yesterday.  oops!  wasn't ignoring you)

> @NickLaMuro I'm leaning towards just ignoring that case / aka let the case that we never call break.

The key part of your argument here is based on "we" (as I read:  the "ManageIQ org"), but this project is called **`activerecord`**`-virtual_attributes`, not **`manageiq`**`-virtual_attributes`, so a project "owned by use" but not "only for us".  To me that gives assumption that this is a project with a some sort of stable API, especially since we are now in a [`1.x`+ version](https://github.com/ManageIQ/activerecord-virtual_attributes/blob/80f8e43d0dd6f4d335140b3a50c1c9f3a3813da1/lib/active_record/virtual_attributes/version.rb#L3).

When you say:

> we own all the callers to this functionality.

You are talking about `github.com/ManageIQ`, not "anyone that could be using this library ever" and I think it is a pretty bad precedence to set that you are will to just breaking functionality on a whim with this project that I see as being designed for others to use.  I am willing to entertain the idea that for the time being we are probably the only users of project (hard to know based off the download numbers from [rubygems.org](https://rubygems.org/gems/activerecord-virtual_attributes)), but I think you and who ever else you consider an owner of this project should determine when this project is no longer considered "beta" enough (if it isn't already) to have a more stable API and ridged depreciation policy.  That, or determine if this really is a project that is intended for others to use, or something we have just broken out (possibly consider a rename if that is the case).

Also, to answer your other question:

> Do we want to ignore this case, or add an escape clause?

I think this would work, but if your end goal is to remove this, I would consider marking this use case as deprecated and remove it in the `2.0` release as well.

kbrock: ok, reving the major version works for me.

The api for this has never been nailed down. So it is hard to know when something changes the public api when we haven't done a good enough job documenting it.

your are right. this should probably be version `0.1.0`

kbrock:
Sorry to change my mind here.

This is just a refactor - it should not be a major version change.
I'll add support for calling virtual_aggregate with `size` which is wrong, but it was allowed before the refactor.

lib/active_record/virtual_attributes/virtual_total.rb
NickLaMuro: Thanks Keenan.  I am good with this solution.

jrafanie: > Preface: I don't have merge rights on this project or really have much of a say in it's direction, so this is mostly an opinion response.

I don't know, your opinion on this was pretty valuable to me.  I agree.  Do we want to get the rest of the open PRs in first or do we want this in with them all and just release a minor version bump?

NickLaMuro: @jrafanie I think what we decided (side conversation) was that the fix he put in for now:

```diff
+          return define_virtual_total(name, relation, options) if method_name == :size
```

Will work for the time being to keep functionality the same, and if we want to consider deprecating this and removing it in 2.0, I think we should follow that up in another PR.

Again, I agree with the direction, just not doing it in a "wild west" manner.

PR #32, opened 2019-06-19
Add support for left outer joins (fix turbo button)

kbrock:
if we use virtual attributes in outer joins, then translate the column

### Overview

In MIQ, the turbo button uses `left_joins` instead of `references`

https://github.com/ManageIQ/manageiq/pull/18892

### Before

- `references`/`includes` properly translate virtual attributes  :+1:
- `left_joins` will not  :-1:

So when using the turbo button, reports blew up.

```
ActiveRecord::ConfigurationError:
    Can't join 'ExtManagementSystem' to association named 'hostname'; perhaps you misspelled it?
 ```

### After

- `references`/`includes` properly translate virtual attributes :+1:
- `left_joins` properly translate virtual attributes :+1:

turbo button works.

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -221,6 +221,12 @@ module ActiveRecord
         super
       end
 
+      # From ActiveRecord::QueryMethods
+      def build_left_outer_joins(manager, outer_joins, *rest)
+        outer_joins = klass.replace_virtual_fields(outer_joins)
+        super if outer_joins.present?
+      end
+
       # From ActiveRecord::Calculations
       def calculate(operation, attribute_name)
         # work around 1 until https://github.com/rails/rails/pull/25304 gets merged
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/active_record/virtual_attributes/virtual_fields.rb [hunk 1]
```diff
@@ -221,6 +221,12 @@ def select(*fields)
         super
       end
 
+      # From ActiveRecord::QueryMethods
+      def build_left_outer_joins(manager, outer_joins, *rest)
+        outer_joins = klass.replace_virtual_fields(outer_joins)
+        super if outer_joins.present?
```
jrafanie: do we need to call `super` regardless?

kbrock: we are modifying `outer_joins`. We could be adding, keeping the same, or removing it.

We only want to call `build_left_outer_joins` if we actually have an `outer_joins` present.

[calling code](https://github.com/rails/rails/blob/58256db50750781016f5ea6d2dd6cbc72096062b/activerecord/lib/active_record/relation/query_methods.rb#L1034) only calls unless `outer_joins.empty?`

PR #33, opened 2019-06-19
Support bye bug

kbrock:
not happy having byebug as a development dependency

but it just became too hard to constantly add this and comment it out

Of Note: Despite what rubocop says, MiQ core includes the byebug gem


kbrock:
yea, the byebugs are only for testing. so ignoring the bot comments

PR #34, opened 2019-06-19
refactor use of deep_merge

kbrock:
Depends:

- [x] #36 first 2 commits

Now that the merge works more inclusively, many of these `case` statements are no longer necessary

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -51,7 +51,11 @@ module ActiveRecord
               merge_includes(h, replace_virtual_fields(virtual_includes(parent)))
             else
               reflection = reflect_on_association(parent.to_sym)
-              new_child = reflection.nil? || reflection.options[:polymorphic] ? {} : reflection.klass.replace_virtual_fields(child) || {}
+              if reflection.nil? || reflection.options[:polymorphic]
+                new_child = {}
+              else
+                new_child = reflection.klass.replace_virtual_fields(child) || {}
+              end
               merge_includes(h, parent => new_child)
             end
           end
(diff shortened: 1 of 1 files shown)


kbrock:
fixed cops

kbrock:
fixed merge conflict in virtual_includes_spec.rb

jrafanie:
> we were getting N+1 errors on the page

what are N+1 errors?



> The result was non deterministic. Sometimes it would pick :hardware from the left and other times from the right

Sounds like a bug...

> I tried to write the refactor as simply as possible

Is there a commit or commits that have only the bug fix?  It's hard to see the bug fix with all the refactoring.  Is the refactoring required to do the fix?  Can you do the refactoring as a separate commit so I can split the bug fix from the refactoring in my mind?

kbrock:
> Is there a commit or commits that have only the bug fix?

I pulled out the first commit into #36 - that is the root of the problem.

> (so if it is 1 commit, why did you introduce 6 in this PR?)

The rest of the commits remove the second `case` statement. This `case` statement is the reason I added 111 `expect`s to the specs. There are so many way to construct a hash, array, symbol tree. Too many edge cases. This new approach reduces the number of cases.

Good news, miq tends to pass all includes as hashes (e.g.: `includes(:hardware => {:disks => {}})`. But the `uses` are defined as symbols, arrays, and hashes. I'm concerned that someone will add a uses or includes clause that we don't yet support. (e.g.: `includes(:hardware => :disks)`). So refactoring this code is a priority for me.

>N+1

```
class VmOrTemplate
  belongs_to :hardware
  virtual_delegate :allocated_disk_storage, :uses => {:hardware => :disks}
end

VmOrTemplate.includes(:hardware).includes(:allocated_disk_storage).map(&:allocated_disk_storage)
```

If the includes gets mapped properly (i.e.: `:hardware => :disks`), the call to `allocated_disk_storage` will have all the data it needs. So all is good.

If the includes doesn't get mapped properly (i.e.: `:hardware`), each call to `allocated_disk_storage` will need to fetch disks for the vm/hardware records. We end up with N+1 records.

PR #35, opened 2019-06-20
fix broken bin/console

kbrock:
the gem was renamed, but the new name was not added to `./bin/console`

This also now connects to an empty test database (as suggested by the bin/console file) so it can be tested and exercised in the console.

### before

```
./bin/console
Traceback (most recent call last):
	1: from ./bin/console:4:in `<main>'
./bin/console:4:in `require': cannot load such file -- virtual_attributes (LoadError)
```

### after

```
./bin/console
>Author.create_with_books(5)
# => #<Author id: 1, name: "foo", nickname: nil>  
>Author.count
# => 1  
>Book.count
# => 5  
```

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -3,12 +3,9 @@
 require "bundler/setup"
 require "active_record-virtual_attributes"
 
-# You can add fixtures and/or initialization code here to make experimenting
-# with your gem easier. You can also use a different console, if you like.
-
-# (If you use this, don't forget to add pry to your Gemfile!)
-# require "pry"
-# Pry.start
+# models for local testing
+require "rspec"
+Dir['./spec/support/**/*.rb'].sort.each { |f| require f }
 
 require "irb"
 IRB.start(__FILE__)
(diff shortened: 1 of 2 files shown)


kbrock:
re ignoring exceptions - I like to allow gem byebug to not be present

spec/support/database.rb
```diff
@@ -1,5 +1,6 @@
 require "logger"
 require "active_record"
+require "erb"
```
carbonin: What is `erb` doing here?

kbrock: https://github.com/ManageIQ/activerecord-virtual_attributes/pull/35/files#diff-08d5f77d01f8f94063c830ce47ebfc74L29

This was throwing an error before without the added `require`

Huh, but looks like it is working now.
So I'll remove it.

... confused ...

kbrock:
Thanks @carbonin 
I removed the require. Still confused why it worked before

PR #36, opened 2019-06-24
virtual includes - better deep merge for attributes

kbrock:
### Bug

We have a performance regression on the VMs page (`VmOrTemplate.yaml`)
Sometimes the page loads objects great, and other times it does an N+1 on `disks`.

This is a subset of #34 

### Overview

The issue stems around with the way we handle `includes()` to load the objects for the VMs page.

```ruby
class VmOrTemplate
  belongs_to :hardware
  virtual_delegate :allocated_disk_storage, :uses => {:hardware => :disks}
end
```

For our views, we generate `includes()` for all virtual columns and associated tables:
- `hardware.name` -- `includes(:hardware)`
- `allocated_disk_storage` -- `includes(:hardware => :disks)` /via `:uses`

```ruby
VmOrTemplate.includes(:hardware).includes(:hardware => :disks).each do |vm|
  puts "#{vm.id} #{vm.hardware.name} #{vm.allocated_disk_storage)
end
```

The issue is when you merge the 2 different nested hashes from the `includes()`.
The algorithm is broken and would generate 1 of 2 possible `includes`:

```ruby
# look up all vms, all hardware, disks for each hardware record (N+2 queries)
VmOrTemplate.includes(:hardware).each do |vm|
  puts "#{vm.id} #{vm.hardware.name} #{vm.allocated_disk_storage)
end

# look up all vms, all hardware, all disks (3 queries)
VmOrTemplate.includes(:hardware => :disks).each do |vm|
  puts "#{vm.id} #{vm.hardware.name} #{vm.allocated_disk_storage)
end
```

### After

`replace_virtual_field_hash` is revamped to produce consistent results.
And bonus, it returns is the hash with all the associations:

`VmOrTemplate.includes(:hardware => :disks)` (3 queries)

This means the N+1 errors on the `VmOrTemplate` page go away.

### DIFF
--- a/spec/virtual_includes_spec.rb
+++ b/spec/virtual_includes_spec.rb
@@ -53,6 +53,10 @@ describe ActiveRecord::VirtualAttributes::VirtualIncludes do
       expect(Author.includes(:nick_or_name => {}, :first_book_name => {})).to preload_values(:first_book_name, book_name)
     end
 
+    it "preloads virtual_attributes dups" do
+      expect(Author.includes(:total_named_books).includes(:first_book_author_name)).to preload_values(:first_book_author_name, author_name)
+    end
+
     it "preloads virtual_attribute (:uses => {:book => :author_name})" do
       expect(Author.includes(:first_book_author_name)).to preload_values(:first_book_author_name, author_name)
       expect(Author.includes([:first_book_author_name])).to preload_values(:first_book_author_name, author_name)
@@ -122,6 +126,10 @@ describe ActiveRecord::VirtualAttributes::VirtualIncludes do
       expect(Author.includes(:books_with_authors => {}, :books => {}).references(:books => {})).to preload_values(:books_with_authors, named_books)
       expect(Author.includes(:books => {}).includes(:books => {:author => {}}).references(:books => {})).to preload_values(:books_with_authors, named_books)
       expect(Author.includes(:books => {:author => {}}).includes(:books => {}).references(:books => {})).to preload_values(:books_with_authors, named_books)
+    end
+
+    it "preloads virtual_attributes dups" do
+      expect(Author.includes(:books => :author, :books_with_authors => {}).references(:first_book_author_name => {})).to preload_values(:first_book_author_name, author_name)
     end
 
     it "preloads virtual_attribute (:uses => {:book => :author_name})" do
@@ -286,6 +294,20 @@ describe ActiveRecord::VirtualAttributes::VirtualIncludes do
     it "merges hash, hash - 1 level only" do
       first  = {:key => {:more1 => {}}}
       second = {:key => {:more2 => {}}}
+      result = {:key => {:more1 => {}, :more2 => {}}}
+      expect(Author.merge_includes(first, second)).to eq(result)
+    end
+
+    it "merges hash(hash), hash(symbol)" do
+      first  = {:key => {:more1 => {}}}
+      second = {:key => :more2}
+      result = {:key => {:more1 => {}, :more2 => {}}}
+      expect(Author.merge_includes(first, second)).to eq(result)
+    end
+
+    it "merges hash(hash), hash(array)" do
+      first  = {:key => {:more1 => {}}}
+      second = {:key => [:more2]}
       result = {:key => {:more1 => {}, :more2 => {}}}
       expect(Author.merge_includes(first, second)).to eq(result)
     end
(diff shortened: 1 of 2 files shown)


kbrock:
fixup rubocops

spec/virtual_includes_spec.rb
```diff
@@ -228,6 +240,64 @@
...
+      first  = {:key => {:more1 => {}}}
+      second = {:key => {:more2 => {}}}
+      result = {:key => {:more1 => {}, :more2 => {}}}
+      expect(Author.merge_includes(first, second)).to eq(result)
+    end
+
+    it "merges 2 overlapping hashes - 2 levels" do
+      first  = {:key => {:more => {:third => {:fourth1 => {}}}}}
+      second = {:key => {:more => {:third => {:fourth2 => true}}}}
+      result = {:key => {:more => {:third => {:fourth1 => {}, :fourth2 => true}}}}
```
jrafanie: Can you add a test like this but where the deeply nested hash keys overlap, to demonstrate who wins?

```
      first  = {:key => {:more => {:third => {:fourth => {}}}}}
      second = {:key => {:more => {:third => {:fourth => true}}}}
      result = {:key => {:more => {:third => {:fourth => ????}}}
```

jrafanie:
LGTM, just added a suggestion for another test example.

spec/virtual_includes_spec.rb
kbrock: 2nd commit added this

spec/virtual_includes_spec.rb
```diff
@@ -228,6 +240,78 @@
...
+      first  = {:key => {:more1 => {}}}
+      second = {:key => {:more2 => {}}}
+      result = {:key => {:more1 => {}, :more2 => {}}}
+      expect(Author.merge_includes(first, second)).to eq(result)
+    end
+
+    it "merges hash(hash), hash(symbol)" do
+      first  = {:key => {:more1 => {}}}
+      second = {:key => :more2}
+      result = {:key => {:more1 => {}, :more2 => {}}}
```
jrafanie: Interesting, so the value of the `:more2` key depends on the value of the`:more1` key?

spec/virtual_includes_spec.rb
```diff
@@ -228,6 +240,78 @@
...
+      expect(Author.merge_includes(first, second)).to eq(result)
+    end
+
+    it "merges hash(hash), hash(array)" do
+      first  = {:key => {:more1 => {}}}
+      second = {:key => [:more2]}
+      result = {:key => {:more1 => {}, :more2 => {}}}
```
jrafanie: Head asplode

jrafanie: It feels like that's not going to be obvious... do you have an example where includes look like this?

spec/virtual_includes_spec.rb
kbrock: We converted `second` from `{:key => :more2}` to `{:key => {:more2 => {}}}`

spec/virtual_includes_spec.rb
kbrock: [Host] has a bunch of different `:uses`:

```ruby
class Host
  virtual_column :v_owning_datacenter, :type => :string, :uses => :all_relationships
  virtual_column :os_image_name,       :type => :string, :uses => [:operating_system, :hardware]
  virtual_has_many :miq_scsi_luns, :uses => {:hardware => {:storage_adapters => {:miq_scsi_targets => :miq_scsi_luns}}}
end
```

It has a key, an array, and a hash pointing to a hash pointing to a hash pointing to a key.

---

[Flavor] shows a few styles of `includes()`

```ruby
class Flavor
  def self.tenant_joins_clause(scope)
    scope.includes(:cloud_tenants => "source_tenant", :ext_management_system => {})
         .references(:cloud_tenants => "source_tenant", :ext_management_system => {})
  end
end
```

It has a hash key pointing to a string and a hash key pointing to a hash.

[Host]: https://github.com/ManageIQ/manageiq/blob/master/app/models/host.rb#L118
[Flavor]: https://github.com/ManageIQ/manageiq/blob/master/app/models/flavor.rb#L46

spec/virtual_includes_spec.rb
kbrock: In our code:

- we mostly pass the second format for `includes()` (the pure hash version).
- we mostly pass the first format for `:uses` (symbols or hashes point to symbols)

spec/virtual_includes_spec.rb
jrafanie: and what does it do if there are multiple values in the array:

this?
```
      first  = {:key => {:more1 => {}}}
      second = {:key => [:more2, :more3]}
      result = {:key => {:more1 => {}, :more2 => {}, :more3 => {}}}
```

jrafanie:
Sorry, was there a bug for this?  I think I follow what's happening here, I just had a minor comment confirming my understanding of what this code is doing.

spec/virtual_includes_spec.rb
kbrock: yup

PR #38, opened 2019-07-12
1.4.0 version notes

kbrock:
thinking this is a minor version since it only updates bugs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ a nice looking [Changelog](http://keepachangelog.com).
 
 ## Version [Unreleased]
 
+## Version [1.4.0] <small>2019-07-13</small>
+
+* fix includes to include all associations
+* fix bin/console to now actually run
+* select no longer munges field attribute
+* support virtual attributes in left_outer_joins
+
 ## Version [1.3.1] <small>2019-06-06</small>
 
 * quote column aliases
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ a nice looking [Changelog](http://keepachangelog.com).
 
 ## Version [Unreleased]
 
+## Version [1.3.2] <small>2019-07-13</small>
+
+* fix includes to include all associations
+* fix bin/console to now actually run
+* select no longer munges field attribute
+* support virtual attributes in left_outer_joins
```
jrafanie: @kbrock are these new functionality or bug fixes?  It sounds like the last might be a never before supported feature, therefore, I think it should be a minor release, 1.4.0, right?

kbrock: @jrafanie includes, bin/console, and select are definitely bug fixes

supporting [left_outer_join](https://github.com/ManageIQ/activerecord-virtual_attributes/pull/32#) is questionable

kbrock:
Going with a 1.4 since left outer join is kind of an enhancement

jrafanie:
> Going with a 1.4 since left outer join is kind of an enhancement

Sounds good. Please update the existing references to 1.3.2 in the changelog, PR title /description.  The branch name is fine.

PR #39, opened 2019-07-17
Support Rails 6.0 polymorphic associations

kbrock:
Depends upon:

- [x] #40 fix for types bug (first 2 commits)
- [x] #42 
- [x] #43 
- [x] #44 / #45 (style fix up per request)
- [x] #46 travis updates
- [x] #50 `select()` updates to work better with `from()`
- [x] #57 `includes()` fix and `:through` fix
---

- When you specify a `from()` clause, rails 5.x stops using `arel_column` and basically breaks virtual attributes. Rails 6.0 works though, so monkey patching 5.{1,2} from rails 6.0.
- Virtual attributes in `select()` are embedded as sub-queries. These need to be accessed in `attributes[]`, and we use an alias for this.
- updates the associations in preload and then lets the standard preload logic work in a standard way
- preloading polymorphic in includes is no longer working. doesnt work in rails either, but we had it working for a little while

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -102,14 +102,27 @@ module ActiveRecord
     class Preloader
       prepend(Module.new {
         if ActiveRecord.version.to_s >= "6.0"
+          def preloaders_for_reflection(reflection, records, scope, polymorphic_parent)
+            case reflection
+            when Array
+              reflection.flat_map { |ref| preloaders_on(ref, records, scope, polymorphic_parent) }
+            when Hash
+              preloaders_on(reflection, records, scope, polymorphic_parent)
+            else
+              super(reflection, records, scope)
+            end
+          end
+
           # rubocop:disable Style/BlockDelimiters, Lint/AmbiguousBlockAssociation, Style/MethodCallWithArgsParentheses
           # preloader.rb active record 6.0
+          # changed:
+          # since grouped_records can return a hash/array, we need to handle those 2 new cases
           def preloaders_for_hash(association, records, scope, polymorphic_parent)
             association.flat_map { |parent, child|
               grouped_records(parent, records, polymorphic_parent).flat_map do |reflection, reflection_records|
-                loaders = preloaders_for_reflection(reflection, reflection_records, scope)
+                loaders = preloaders_for_reflection(reflection, reflection_records, scope, polymorphic_parent)
                 recs = loaders.flat_map(&:preloaded_records).uniq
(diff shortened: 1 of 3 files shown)


jrafanie:
I had to make a change to the monkey patch for 5.2 support, you can follow along here:  https://github.com/juliancheal/activerecord-virtual_attributes/commits/try_rails_5_2

There's still lots of test failures I hope to dig into tomorrow.

kbrock:
> Method arel_column has a Cognitive Complexity of 12 (exceeds 11 allowed). Consider refactoring

it was 90% rails monkey patch

kbrock:
@jrafanie @juliancheal This passes miq with 51 and 52 here.
Couldn't get an miq 52 rails branch to bundle - let me know which branch to use over there.
Also let me know if this resolves most of your problems.

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -103,20 +103,11 @@ class Base
...
-
-          records_with_association = klass_map.select { |k, _rs| k.reflect_on_association(association) }.flat_map { |_k, rs| rs }
-          if records_with_association.any?
-            loaders.concat(super(association, records_with_association, scope))
-          end
-
-          loaders
+        def preload(records, associations, preload_scope = nil)
+          records = Array.wrap(records).compact
+          associations = records.first.class.replace_virtual_fields(associations) if records.first
```
kbrock: only associations from the first class will be used

can we use `group_by` / `grouped_records` ?

kbrock: also, did polymorphic work before us?
how much of polymorphic did we loose?

Gemfile
```diff
@@ -2,9 +2,9 @@
 
 source "https://rubygems.org"
 
-gem "activerecord", "~> 5.0.7"
-gem "mysql2", '~> 0.4.0'
+gem "activerecord", "~> 5.2.3"
```
kbrock: we are modifying the gemfile to be more current

Also was concerned nailing down a gem version when future versions of active record no longer require a particular version

lib/active_record/virtual_attributes/virtual_fields.rb
kbrock: added grouping by class.

kbrock:
updated and removed 6.0 support

kbrock:
kicking

kbrock:
This was a PR for rails 5.2. It became a PR to support rails 6.0

There is 1 remaining habtm preload failure with rails 5.2. (but rails 5.2 is doing pretty well)

juliancheal:
🎉 YAY!

PR #42, opened 2019-09-11
use references properly

kbrock:
Rails uses `references(table_names)`
and knows nothing about hashes or relations 
(nor has it ever known).

These tests fix that problem.

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -152,7 +152,7 @@ module ActiveRecord
         # This can be seen with the following:
         #
         #   Vm.select(Vm.arel_table[Arel.star]).select(:some_vm_virtual_col)
-        #     .includes(:tags => {}).references(:tags => {})
+        #     .includes(:tags => {}).references(:tags)
         #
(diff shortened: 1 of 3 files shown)


jrafanie:
rebase please @kbrock

kbrock:
> Can you reference either in the commit or PR description where this exists or was changed in Rails. I am almost positive that it used to work this way in Rails, and now has since changed.
> - @NickLaMuro 

The method was [added in 4.0](https://github.com/rails/rails/commit/4c4760a619e9bddf14e65dd7f0d5bbc3f9ca320e) and has always been just table names

The `to_s` was [added 3 months later](https://github.com/rails/rails/commit/8c2c60511bea) and has been in there ever since. So a hash would never have worked.

the api also clearly states it is table names not relation names.

There is an escape clause, that says just `references` everything if the table is not in the includes. so `references("x")` is just as effective.

> Also, pretty sure we have a case or two in our code base.

I remove these invalid `references()` in https://github.com/ManageIQ/manageiq/pull/19127
It may make sense to handle the static and dynamic cases (i.e.: reporting) in separate PRs.

jrafanie:
LGTM

PR #43, opened 2019-09-11
specs for preloaded

kbrock:
we will refactoring the way preload works.
So we want specs in place.

We already had specs for `includes()` and `includes().references()`.

This PR adds `preload` specs.


kbrock:
those exceptions come back in future PR

PR #44, opened 2019-09-30
Version specific versions

kbrock:
Class names change across different active record versions.

This adds case statements to handle these types of changes across active record 5.2 and 6.0

This is extracted out of #39

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -183,8 +183,11 @@ module ActiveRecord
           additional_attributes = result_set.first.keys
                                             .reject { |k| join_dep_keys.include?(k) }
                                             .reject { |k| join_root_aliases.include?(k) }
-                                            .map    { |k| [k, k] }
-          column_aliases += additional_attributes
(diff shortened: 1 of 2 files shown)


kbrock:
The complaint is that the monkey patch to rails is too complex.

sorry RuboCop. this is complex stuff

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -183,8 +183,11 @@ def instantiate(result_set, *_, &block)
           additional_attributes = result_set.first.keys
                                             .reject { |k| join_dep_keys.include?(k) }
                                             .reject { |k| join_root_aliases.include?(k) }
-                                            .map    { |k| [k, k] }
-          column_aliases += additional_attributes
+          column_aliases += if ActiveRecord.version.to_s < "6.0"
```
jrafanie: @kbrock I'm good with this but looking at this PR I'm wondering if we should always use the same format for these version checks...  in other words flip this from `<` to `>=` and flip the if/else.  If you agree, maybe you can do this in a followup PR.  It's the only suggestion I can make.

PR #46, opened 2019-10-01
Travis and Appraisal changes

kbrock:
- Added definitions for active record 5.2 and 6.0
- Upgraded gemfiles to be more similar to other projects.
- Alphabetized the list of services and gems
- updated the versions of dependent gems


.travis.yml
```diff
@@ -1,20 +1,26 @@
 ---
+sudo: false
 language: ruby
 cache: bundler
 rvm:
-- 2.4.1
-- 2.5.3
+- 2.4.6
```
NickLaMuro: > test against latest versions of gems 

... and rubies :trollface:

.travis.yml
```diff
@@ -13,14 +13,14 @@ env:
 - DB=pg
 - DB=sqlite3
 gemfile:
-- gemfiles/virtual_attributes_50.gemfile
-- gemfiles/virtual_attributes_51.gemfile
-#- gemfiles/virtual_attributes_52.gemfile
-#- gemfiles/virtual_attributes_60.gemfile
+- gemfiles/gemfile_50.gemfile
```
NickLaMuro: No where to put this, but there is a typo in your commit message:

> So it is easier ff they all are basically the same with the same file
> names.

`s/ff/if/`

.travis.yml
kbrock: sigh... thanks

kbrock: well, feels like this has been in the works since 2.5.3 was the latest

NickLaMuro: @kbrock I was commenting on the fact that your commit message said "test against latest versions of gems", but you are also updating the versions of ruby you are testing against.

Again, just trolling.

PR #47, opened 2019-10-02
Put version check around calculate monkey patch

kbrock:
In code, mark this monkey patch as an pre ActiveRecord 5.0 only patch.

1. Make the version of active record that we are patching more explicit
2. Add tests to properly test this code.
3. Remove code that does the same conversion twice

- [x] https://travis-ci.org/kbrock/manageiq-cross_repo/builds/596266507


kbrock:
I'm happy with the new changes:

- added tests for `includes().count` (very common case and made controllers fail)
- added tests for `sum(virtual_attribute)` (wasn't tested)
- put version check around `calculate` tests - detected problem with my first bad monkey patch
- removed duplicate code that converted virtual attribute to arel.

PR #49, opened 2019-10-07
[WIP] stubs to sanity check queries

kbrock:
`STUBS=true bundle exec rspec/simple_spec.rb`

this shows how things work using the model without virtual attributes.

You can also go into the `bin/console` to see how the models behave with default rails behaviors


kbrock:
WIP: I will keep using this
but not sure if it belongs in the main repo

kbrock:
looks like this will just live locally in my little world

PR #52, opened 2019-10-16
Rewrite virtual_total attribute_arel

kbrock:
There are a number of places in miq where using a `virtual_total` would simplify our code base. But those spots required that `virtual_total` works with `has_many :through`

This adds support for a `virtual_total` with a relation that uses one or more `:through => model`.

### before

manually construct the arel for `virtual_total` attributes.
This required that we know a lot about scopes, `join_keys`, and other edge cases.

We also were not able to handle the edge case of having `:through => :model`

### after

Let Active Record construct the arel for `virtual_total` attributes. We then tweak the arel by remove the primary table from the sub query since the table is already in our primary query.

We now support any relation that rails supports, since we are letting rails do the work for us

- [ ] https://travis-ci.org/kbrock/manageiq-cross_repo/builds/600719235
- [ ] https://github.com/ManageIQ/manageiq/pull/19410 fix manageiq

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_total.rb
+++ b/lib/active_record/virtual_attributes/virtual_total.rb
@@ -99,15 +99,18 @@ module VirtualAttributes
 
       def virtual_aggregate_arel(reflection, method_name, column)
         return unless reflection && reflection.macro == :has_many && !reflection.options[:through]
+
+        # need db access for the reflection join_keys, so delaying all this key lookup until call time
         lambda do |t|
+          # query: SELECT * FROM foreign_table
           query = if reflection.scope
                     reflection.klass.instance_exec(nil, &reflection.scope)
                   else
                     reflection.klass.all
                   end
 
+          # query: SELECT * FROM foreign_table [WHERE main_table.id = foreign_table.id]
           foreign_table = reflection.klass.arel_table
-          # need db access for the keys, so delaying all this lookup until call time
           if ActiveRecord.version.to_s >= "5.1"
             join_keys = reflection.join_keys
           else
@@ -115,6 +118,8 @@ module VirtualAttributes
           end
           query       = query.except(:order).where(t[join_keys.foreign_key].eq(foreign_table[join_keys.key]))
 
+
+          # query: [SELECT COUNT(*)] FROM foreign_table WHERE main_table.id = foreign_table.id
           arel_column = if method_name == :size
                           Arel.star.count
                         else
@@ -122,7 +127,10 @@ module VirtualAttributes
                         end
           query       = query.select(arel_column)
 
-          t.grouping(Arel::Nodes::SqlLiteral.new(query.to_sql))
+          sql         = query.to_sql
+
+          # add () around query
+          t.grouping(Arel::Nodes::SqlLiteral.new(sql))
         end
       end
     end
--- a/lib/active_record/virtual_attributes/virtual_total.rb
+++ b/lib/active_record/virtual_attributes/virtual_total.rb
@@ -102,6 +102,13 @@ module VirtualAttributes
 
         # need db access for the reflection join_keys, so delaying all this key lookup until call time
         lambda do |t|
+          # arel_column: COUNT(*)
+          arel_column = if method_name == :size
+                          Arel.star.count
+                        else
+                          reflection.klass.arel_attribute(column).send(method_name)
+                        end
+
           # query: SELECT * FROM foreign_table
           query = if reflection.scope
                     reflection.klass.instance_exec(nil, &reflection.scope)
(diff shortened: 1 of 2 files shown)


lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -98,31 +98,36 @@ def define_virtual_aggregate_method(name, relation, method_name, column)
...
-          arel_column = if method_name == :size
-                          Arel.star.count
-                        else
-                          reflection.klass.arel_attribute(column).send(method_name)
-                        end
-          query       = query.select(arel_column)
-
-          t.grouping(Arel::Nodes::SqlLiteral.new(query.to_sql))
+          # arel_column: COUNT(*)
+          arel_column = (column == Arel.star ? column : reflection.klass.arel_attribute(column)).send(method_name)
```
kbrock: I wanted to call arel_attribute if column is a string or symbol, but `Arel.star.kind_of?(String)`

The goal of using `Arel.star` is to avoid the hardcoded logic here. (especially basing the column upon the method field)

skipping the `if` check essentially calls `reflection.klass.arel_attribute(Arel.star)` - which will call `count(table.*)` which is good on pg but fails in sqlite and mysql - and is not really what I was aiming for.

I'm very happy how passing `method_name` as `:count` works now. much better than passing `:size` and changing to `:count` here.

BUT this change is not necessary (it is the last commit) and I could remove it from this PR.
I just through since we are focusing on testing the counts here, it would make sense to have both changes together.

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -98,31 +98,36 @@ def define_virtual_aggregate_method(name, relation, method_name, column)
...
+          query            = relation_query.arel
+          bound_attributes = relation_query.bound_attributes
+
+          # algorithm:
+          # - remove main_table from this sub query. (it is already in the primary query)
+          # - move the foreign_table from the JOIN to the FROM clause
+          # - move the main_table.id = foreign_table.id from the ON clause to the WHERE clause
+
+          # query: SELECT COUNT(*) FROM main_table [ ] JOIN ...
+          join = query.source.right.shift
```
kbrock: I had wanted to comment here what was in `join` but I wasn't able to read it.

I outlined the algorithm a few lines above, and this basically just shows the algorithm working. But both together made it a little easier for me to follow.

Do I need to better explain WHY we are removing the `main_table` from the sub query that we are building?
When you see `sub query` is it obvious that this is the query we're building and the `main query` is actually an abstract bigger thing that is not part of this method?

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -109,26 +109,26 @@ def virtual_aggregate_arel(reflection, method_name, column)
...
+          # query: SELECT COUNT(*) FROM [foreign_table] JOIN ...
+          query.source.left = join.left
+          # query: SELECT COUNT(*) FROM foreign_table JOIN ... [WHERE main_table.id = foreign_table.id]
+          query.where(join.right.expr)
+
+          # convert bind variables from ? to actual values. otherwise, sql is incomplete
+          conn = connection
+          sql  = conn.unprepared_statement { conn.to_sql(query, bound_attributes) }
```
jrafanie: naive question... are `unprepared_statement` and `bound_attributes` public methods and work the same with rails 5.1, 5.2 and 6.0?

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -98,31 +98,40 @@ def define_virtual_aggregate_method(name, relation, method_name, column)
...
-          # need db access for the keys, so delaying all this lookup until call time
-          if ActiveRecord.version.to_s >= "5.1"
-            join_keys = reflection.join_keys
-          else
-            join_keys = reflection.join_keys(reflection.klass)
-          end
-          query       = query.except(:order).where(t[join_keys.foreign_key].eq(foreign_table[join_keys.key]))
+        return unless reflection && reflection.macro == :has_many
 
+        # need db access for the reflection join_keys, so delaying all this key lookup until call time
```
jrafanie: ❤️ Delayed `connection` calls are awesome...

jrafanie:
> * [ ]  https://travis-ci.org/kbrock/manageiq-cross_repo/builds/598861713
> * [ ]  [ManageIQ/manageiq#19410](https://github.com/ManageIQ/manageiq/pull/19410) fix manageiq

@kbrock does this need cross repo run?  The first link is broken on miq.

Does the second one depend on this or does this PR depend on the second one?

kbrock:
> ManageIQ/manageiq#19410 fix manageiq

That PR depends upon this.
That PR said an attribute is not sql friendly. This PR made the attribute sql friendly. So I changed that test to now know that the capabilities of the system have been improved

I updated the cross repo so it will test miq with that pr

lib/active_record/virtual_attributes/virtual_total.rb
kbrock: fixed

lib/active_record/virtual_attributes/virtual_total.rb
kbrock: This is taken from ActiveRecord `Relation#to_sql` as the official way to bind variables back into sql statements.

The `to_sql` statement (if you ignore moving code around) has been the same since before 4.0 and remains the same through 6.0.

thanks to @NickLaMuro for pointing me in the right direction here

PR #53, opened 2019-10-17
No longer hard code size method in arel generation

kbrock:
Creating the aggregate attribute arel has 2 paths: one for `size` and one for all others.
So there is no need to pass in a `method=size` and then later change the value to `count`.
Instead, passing in `method=count` and remove the `if` check.

The column itself is simpler. If it is an arel friendly column,
leave it as arel (checking `respond_to?(:count)`), otherwise convert
to an arel attribute


kbrock:
this conflicted after merging 52, now it is all set again

PR #54, opened 2019-11-12
virtual_total refactorings

kbrock:
This is pulled out of #15 

When attempting to come up with an agreement between #13 and #15 there are some common changes made.

These are the changes from that PR (well both to be honest) that do not change any results.


kbrock:
Not sure what code climate is complaining about. It is much more less complex to me
I will fix this issue later

PR #57, opened 2019-11-25
Rails 52 includes

kbrock:
find_with_associations is only available for 5.1

virtual_arel with the `to_sql` is new from the `:through` fix. but easy enough to fix

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -225,12 +225,23 @@ module ActiveRecord
 
     include(Module.new {
       # From ActiveRecord::FinderMethods
-      def find_with_associations(&block)
-        real = without_virtual_includes
-        if real.equal?(self)
-          super
-        else
-          real.find_with_associations(&block)
+      if ActiveRecord.version.to_s >= "5.2"
+        def apply_join_dependency(*args, &block)
+          real = without_virtual_includes
+          if real.equal?(self)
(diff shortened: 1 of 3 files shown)


lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -225,12 +225,23 @@ def without_virtual_includes
 
     include(Module.new {
       # From ActiveRecord::FinderMethods
-      def find_with_associations(&block)
-        real = without_virtual_includes
-        if real.equal?(self)
-          super
-        else
-          real.find_with_associations(&block)
+      if ActiveRecord.version.to_s >= "5.2"
```
jrafanie: Do we have a way to test this against 5.2 in specs?

jrafanie:
LGTM, but I wasn't sure if we already have tests for 5.2 or need them.

lib/active_record/virtual_attributes/virtual_fields.rb
kbrock: 5.2 got a 3 new failures. these are 2 of them.

Do you want me to enable 5.2?
I can enable them, but they will fail until the 3rd failure is addressed.

jrafanie: ah, yeah, can you enable them with 5.2 to demonstrate it's fixed and maybe skip the test on 5.2 for the expected failure?

PR #58, opened 2019-11-25
separate instantiate with monkey patch to simplify

kbrock:
code climate is saying the code complexity is 14 (exceeds 11 allowed)

simply pulling out code into another method

removing a few RuboCop overrides that are no longer necessary after this change

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -140,7 +140,7 @@ module ActiveRecord
         column_aliases = aliases.column_aliases(join_root)
 
         # New Code
-        column_aliases += select_values_from_references(column_aliases, result_set) unless result_set.empty?
+        column_aliases += select_values_from_references(column_aliases, result_set) if result_set.present?
         # End of New Code
 
         message_bus = ActiveSupport::Notifications.instrumenter
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -191,7 +191,7 @@ module ActiveRecord
       #
       # This is because rails is trying to reduce the number of queries
       # needed to fetch all of the records in the include, so it grabs the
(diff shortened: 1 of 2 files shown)


lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -140,51 +140,7 @@ def instantiate(result_set, *_, &block)
...
-          additional_attributes = result_set.first.keys
-                                            .reject { |k| join_dep_keys.include?(k) }
-                                            .reject { |k| join_root_aliases.include?(k) }
-          column_aliases += if ActiveRecord.version.to_s >= "6.0"
-                              additional_attributes.map { |k| Aliases::Column.new(k, k) }
-                            else
-                              additional_attributes.map { |k| [k, k] }
-                            end
-        end
+        column_aliases += select_values_from_references(column_aliases, result_set) unless result_set.empty?
```
djberg96: I think `if result_set.present?` is the currently favored parlance.

kbrock: since `result_set` should be `loaded?`, I suppose that the 2 would be the same thing.

Changes like this make me nervous since `empty?`, `blank?`, `present?` all seem to be slightly different in the active record land.

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -209,7 +165,53 @@ def instantiate(result_set, *_, &block)
...
+      #          "vms"."vendor"  AS t0_r1
+      #          "vms"."format"  AS t0_r1
+      #          "vms"."version" AS t0_r1
+      #          ...
+      #          "tags"."id"     AS t1_r0
+      #          "tags"."name"   AS t1_r1
+      #
+      # This is because rails is trying to reduce the number of queries
+      # needed to fetch all of the records in the include, so it grabs the
+      # columns for both of the tables together to do it.  Unfortuantely (or
```
djberg96: "Unfortunately"

kbrock:
that unfortunate typo was from the original code, but fortunately I fixed it up

PR #63, opened 2020-03-31
work around attribute.name [5.2, 6.0]

kbrock:
Rails is now calling:
```
attribute = arel_attribute()
attribute.name
```

All of our `arel_attributes` return a grouping or an attribute

This change allows our return values from `arel_attributes`
to respond to `grouping.name`, so serialization information can be looked
up in rails.

I looked into deferring the call to `arel_attributes` in rails, but this
requires a change in the attribute handlers, which has been an exposed
interface for a while, so it is not possible.
I'm leaving as monkey patch stage.

We may want to look into getting closer to the attribute code

Not thrilled with this solution.
Can't subclass Grouping because class is hardcoded.
Can't subclass Attribute for same reason.

introduced by https://github.com/rails/rails/pull/32981
cross repo: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/105

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_arel.rb
+++ b/lib/active_record/virtual_attributes/virtual_arel.rb
@@ -9,6 +9,10 @@ module ActiveRecord
     #
     # Model.select(Model.arel_table.grouping(Model.arel_table[:field2]).as(:field))
     # Model.attribute_supported_by_sql?(:field) # => true
+    class Arel::Nodes::Grouping
+      attr_accessor :name
+    end
+
     module VirtualArel
       extend ActiveSupport::Concern
 
@@ -21,8 +25,11 @@ module ActiveRecord
         def arel_attribute(column_name, arel_table = self.arel_table)
           load_schema
           if virtual_attribute?(column_name) && !attribute_alias?(column_name)
-            col = _virtual_arel[column_name.to_s]
-            col.call(arel_table) if col
+            if (col = _virtual_arel[column_name.to_s])
+              arel = col.call(arel_table)
+              arel.name = column_name if arel.kind_of?(Arel::Nodes::Grouping)
+              arel
+            end
           else
             super
           end
--- a/spec/db/models.rb
+++ b/spec/db/models.rb
@@ -51,9 +51,9 @@ class Author < VirtualTotalTestBase
   end
 
   # a (local) virtual_attribute without a uses, but with arel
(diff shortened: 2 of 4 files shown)


lib/active_record/virtual_attributes/virtual_arel.rb
```diff
@@ -9,6 +9,10 @@ module VirtualAttributes
     #
     # Model.select(Model.arel_table.grouping(Model.arel_table[:field2]).as(:field))
     # Model.attribute_supported_by_sql?(:field) # => true
+    class Arel::Nodes::Grouping
```
jrafanie: Should the comments above  be moved down, with VirtualArel?

lib/active_record/virtual_attributes/virtual_arel.rb
```diff
@@ -21,8 +25,11 @@ module ClassMethods
         def arel_attribute(column_name, arel_table = self.arel_table)
           load_schema
           if virtual_attribute?(column_name) && !attribute_alias?(column_name)
-            col = _virtual_arel[column_name.to_s]
-            col.call(arel_table) if col
+            if (col = _virtual_arel[column_name.to_s])
+              arel = col.call(arel_table)
+              arel.name = column_name if arel.kind_of?(Arel::Nodes::Grouping)
```
jrafanie: Can we initialize the name when Arel::Nodes::Grouping is created?  Why is this type different?

spec/db/models.rb
```diff
@@ -51,9 +51,9 @@ def nick_or_name
   end
 
   # a (local) virtual_attribute without a uses, but with arel
-  virtual_attribute :nick_or_name, :string do |t|
+  virtual_attribute :nick_or_name, :string, :arel => (lambda do |t|
```
jrafanie: Interesting, we were calling `.call` previously... was this a bug previously or something else?

kbrock: this was a bug - it wasn't defining the arel (and caused errors in my new tests)

lib/active_record/virtual_attributes/virtual_arel.rb
kbrock: the `call` is calling into every attribute defined.
the name is part of the general arel interface

so, unfortunately, this needs to be here (and is a hack)

jrafanie:
The cross repo tests pass on all but a network error for manageiq-api.

PR #64, opened 2020-03-31
Version check for grouping monkey patch

kbrock:
We had a temporary patch for grouping while a pr was merged
It has been incorporated in rails 6.0, so updating the patch
to let us know this and when we can delete it

fixed in: https://github.com/rails/arel/pull/449

travis was not linking.
change: https://travis-ci.org/github/ManageIQ/activerecord-virtual_attributes/builds/669392260

since other are not running 6.0, no need for cross repo


kbrock:
kicking

jrafanie:
> It has been incorporated in rails 6.0
can you include the link to the change that was included in rails 6.0?

PR #67, opened 2020-04-22
Display warnings for invalid attributes

kbrock:
Display warnings for invalid attributes

before:
we were throwing errors, which is correct, but introducing this error checking causes
too many problems.

after:
The deprecation will give us time to fix the issues
in a more reasonable time period.

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -190,8 +190,10 @@ module ActiveRecord
 
               case association
               when Symbol, String
+                # 4/24/20 we want to revert #67 once we handle all these error cases in our codebase.
                 reflection = record.class._reflect_on_association(association)
-                next if polymorphic_parent && !reflection || !record.association(association).klass
+                display_virtual_attribute_deprecation("#{record.class.name}.#{association} does not exist") if !reflection && !polymorphic_parent
+                next if !reflection || !record.association(association).klass
               when nil
                 next
               else # need parent (preloaders_for_{hash,one}) to handle this Array/Hash
@@ -202,6 +204,21 @@ module ActiveRecord
             h
           end
           # rubocop:enable Style/BlockDelimiters, Lint/AmbiguousBlockAssociation, Style/MethodCallWithArgsParentheses
+
+          def display_virtual_attribute_deprecation(str)
+            short_caller = caller
+            # if debugging is turned on, don't prune the backtrace.
+            # if debuggint is off, prune down to the line where the sql is executed
+            # this defaults to false and only displays 1 line number.
+            unless ActiveSupport::Deprecation.debug
+              bc = ActiveSupport::BacktraceCleaner.new
+              bc.add_silencer { |line| line =~ /virtual_fields/ }
+              bc.add_silencer { |line| line =~ /active_record/ }
+              short_caller = bc.clean(caller)
+            end
+
+            ActiveSupport::Deprecation.warn(str, short_caller)
+          end
         else
           def preloaders_for_one(association, records, scope)
             klass_map = records.compact.group_by(&:class)
--- a/spec/virtual_includes_spec.rb
+++ b/spec/virtual_includes_spec.rb
@@ -297,13 +297,20 @@ RSpec.describe ActiveRecord::VirtualAttributes::VirtualIncludes do
     end
 
     if ActiveRecord.version.to_s >= "5.2"
-      # Note: this is non-standard 5.2. 5.2 has a bug that eats these invalid references. This code handles that case correctly
-      it "catches errors" do
-        expect { Author.includes(:invalid).load }.to raise_error(ActiveRecord::ConfigurationError)
-        expect { Author.includes(:books => :invalid).load }.to raise_error(ActiveRecord::ConfigurationError)
+      # NOTE: ActiveSupport::Deprecation.debug = true will show the whole backtrace
+      it "deprecates invalid include" do
+        expect do
+          Author.preload(:invalid).load
+        end.to output(/DEPRECATION WARNING.*virtual_includes_spec/).to_stderr
+      end
+
+      it "deprecates invalid nested includes" do
+        expect do
+          Author.preload(:books => :invalid).load
+        end.to output(/DEPRECATION WARNING.*virtual_includes_spec/).to_stderr
       end
     else
-      it "ignores errors" do
+      it "ignores invalid includes" do
         expect { Author.includes(:invalid).load }.not_to raise_error
         expect { Author.includes(:books => :invalid).load }.not_to raise_error
       end


kbrock:
fixed cops

kbrock:
addressing cops and code climate

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -202,6 +205,20 @@ def grouped_records(orig_association, records, polymorphic_parent)
             h
           end
           # rubocop:enable Style/BlockDelimiters, Lint/AmbiguousBlockAssociation, Style/MethodCallWithArgsParentheses
+
+          def display_virtual_attribute_deprecation(str, skip)
+            return if skip
+
+            short_caller = caller
+            unless ActiveSupport::Deprecation.debug
```
NickLaMuro, line 213: This seems backwards.  Shouldn't this be `if ActiveSupport::Deprecation.debug`, then we add the backtrace?

spec/virtual_includes_spec.rb
```diff
@@ -297,10 +297,17 @@
     end
 
     if ActiveRecord.version.to_s >= "5.2"
-      # Note: this is non-standard 5.2. 5.2 has a bug that eats these invalid references. This code handles that case correctly
-      it "catches errors" do
-        expect { Author.includes(:invalid).load }.to raise_error(ActiveRecord::ConfigurationError)
-        expect { Author.includes(:books => :invalid).load }.to raise_error(ActiveRecord::ConfigurationError)
+      # NOTE: ActiveSupport::Deprecation.debug = true will show the whole backtrace
+      it "supports preloading of invalid association" do
```
NickLaMuro, line 301: Can this description be be similar to the spec below it?

```ruby
it "deprecates invalid includes" do
end

it "deprecates invalid nested includes" do
end
```

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -191,7 +191,10 @@ def grouped_records(orig_association, records, polymorphic_parent)
               case association
               when Symbol, String
                 reflection = record.class._reflect_on_association(association)
-                next if polymorphic_parent && !reflection || !record.association(association).klass
+                if !reflection || !record.association(association).klass
```
NickLaMuro, line 194: Based on the previous `if`, I feel like it is safer to just keep `polymorphic_parent` in it as is:

```ruby
if polymorphic_parent && !reflection || !record.association(association).klass
```

Since it expands to this:

```
if (polymorphic_parent && !reflection) || !record.association(association).klass
```

Where now, `!reflection` isn't dependent of `polymorphic_parent` being in place.

I might not matter, but I think for this being a minor update, I would rather have the least amount of code/logic change to get the desired effect.

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -191,7 +191,10 @@ def grouped_records(orig_association, records, polymorphic_parent)
...
+                  display_virtual_attribute_deprecation("#{record.class.name}.#{association} does not exist", polymorphic_parent)
```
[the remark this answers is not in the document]
kbrock: yea, I like/had the `unless polymorphic` but rubocop was complaining that I had too many ifs. I'll move back and find another way

lib/active_record/virtual_attributes/virtual_fields.rb
kbrock: debug says "please print a long stack trace" so in that instance we don't want to filter the backtrace down.

I wanted the backtrace to point people to the spot in the code where they used the bad include.

lib/active_record/virtual_attributes/virtual_fields.rb
kbrock: if we did not want to display a deprecation warning, then we could just modify the `if`.

but we have 3 cases:

1. association is found. include it.
2. association is not found but it has a polymorphic parent. just skip it.
3. association is not found and parent is not polymorphic. warn and skip (this changed)

We were following the rails 6.0 logic. now we are introducing leniency in case 3.

lib/active_record/virtual_attributes/virtual_fields.rb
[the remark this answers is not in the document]
kbrock: I listened to rubocop and it was a mistake.

lib/active_record/virtual_attributes/virtual_fields.rb
kbrock: I put in a comment to explain

lib/active_record/virtual_attributes/virtual_fields.rb
NickLaMuro: > but we have 3 cases:

But the `if` isn't 3 cases.  None of those cases account for the `!reflection` portion of the statement, and it's original form, that is dependent on `polymorphic_parent`.

In the "new form", it is different, since `polymorphic_parent` now applies to both `!reflection` and the `!record.association(association).klass` case.

lib/active_record/virtual_attributes/virtual_fields.rb
NickLaMuro: I see.  We always print a stacktrace, but how "big" said trace is is based on this.

Fair enough.

kbrock: the default case is printing 1 line. which it printed a line in virtual attributes which was no help.
now it points the the ruby that called the query. not perfect, but the best we can do. (shortcoming of active record itself)

lib/active_record/virtual_attributes/virtual_fields.rb
NickLaMuro: DISREGARD!

I forgot that we were blowing up before, and now we are trying to not do that as part of this change.  Derp.

That said, do we eventually revert to the old if then?

kbrock: Probably best to have semantic errors throw errors.

do this as a todo?

NickLaMuro: > Probably best to have semantic errors throw errors.

Huh?  I don't understand what this means.

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -202,6 +204,21 @@ def grouped_records(orig_association, records, polymorphic_parent)
             h
           end
           # rubocop:enable Style/BlockDelimiters, Lint/AmbiguousBlockAssociation, Style/MethodCallWithArgsParentheses
+
+          def display_virtual_attribute_deprecation(str)
+            short_caller = caller
+            # if debugging is turned on, don't prune the backtrace.
+            # if debuggint is off, prune down to the line where the sql is executed
```
jrafanie, line 211: typo: debugging (followup PR please)

PR #71, opened 2020-05-21
no longer support invalid virtual_column calls

### LINKED ISSUE #70: Deprecation warnings on ruby 2.7
When running with ruby 2.7 there are a large number of deprecation warnings 
```
/home/grare/adam/src/manageiq/manageiq/app/models/manageiq/providers/cloud_manager/vm.rb:41: warning: Passing the keyword argument as the last hash parameter is deprecated
/home/grare/adam/.gems/2.7.0/gems/activerecord-virtual_attributes-1.6.0/lib/active_record/virtual_attributes.rb:51: warning: The called method `virtual_column' is defined here
```

https://github.com/ManageIQ/manageiq/issues/19678

kbrock:
The `virtual_column` method was sometimes called with a type instead of `:type =>`.
While this was never valid, it was prevalent enough to code a fix.

These have all been cleared up, and this PR officially drops support for the bad call.

we're dropping support because ruby 2.7 doesn't like
a hash and named keys right next to each other.

This removes the warning:

```
warning: Passing the keyword argument as the last hash parameter is deprecated
```

Fixes #70 

/cc @agrare https://github.com/ManageIQ/manageiq/issues/19678

### DIFF
--- a/lib/active_record/virtual_attributes.rb
+++ b/lib/active_record/virtual_attributes.rb
@@ -48,13 +48,9 @@ module ActiveRecord
       #
 
       # Compatibility method: `virtual_attribute` is a more accurate name
-      def virtual_column(name, type_or_options, **options)
-        if type_or_options.kind_of?(Hash)
-          options = options.merge(type_or_options)
-          type = options.delete(:type)
-        else
-          type = type_or_options
-        end
+      def virtual_column(name, **options)
+        type = options.delete(:type)
+        raise ArgumentError, "missing :type attribute" unless type
 
         virtual_attribute(name, type, **options)
       end


jrafanie:
should we deprecate this first or was it already deprecated?

kbrock:
> should we deprecate this first or was it already deprecated?
> @jrafanie 

it was never valid. It was just easier to add the support and then fix the instances.

The issue is the signatures for `virtual_attribute` and `virtual_column` are similar

```ruby
  virtual_attribute :host_name, :string, :uses => "Host"
  virtual_column :host_name, :type => :string, :uses => "Host"
```

Those do the same exact thing. Some people introduced `virtual_column`s with the wrong signature.

I looked through all our repos and we no longer have bad definitions. I went through and fixed them all a few years ago.

PR #74, opened 2020-07-24
Update preload_values.rb

d-m-u:
s/it/in and some super trivial dogmatic comma placement nits


kbrock:
I'll merge only if you promise to fix this class to use the nifty matchers as well (and remove QueryCount class)

PR #75, opened 2020-07-24
remove QueryCounter

d-m-u:
as promised, we can use the query counter from db-query-matchers
entirely too easy.

https://github.com/ManageIQ/activerecord-virtual_attributes/pull/74#issuecomment-663678993


kbrock:
Thanks. Nice to get rid of some of this extra code.

PR #76, opened 2020-08-20
ensure various ordering option works

kbrock:
add specs to ensure the various ways to define ordering work with virtual attributes

### DIFF
--- a/spec/virtual_attributes_spec.rb
+++ b/spec/virtual_attributes_spec.rb
@@ -816,16 +816,38 @@ RSpec.describe ActiveRecord::VirtualAttributes::VirtualFields do
   end
 
   describe ".order" do
-    it "supports virtual attributes symbol" do
-      Author.order(:nick_or_name).first # fails
+    before do
+      Author.delete_all
+      authors
     end
 
-    it "supports virtual attributes hash" do
-      Author.order(:nick_or_name => "ASC").first # fails
+    let(:authors) do
+      [
+        Author.create(:name => "aaa"),
+        Author.create(:nickname => "bbb"),
+        Author.create(:name => "ccc")
+      ]
     end
 
-    it "supports virtual attributes arel" do
-      Author.order(Author.arel_attribute(:nick_or_name)).first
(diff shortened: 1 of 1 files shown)


spec/virtual_attributes_spec.rb
```diff
@@ -816,17 +816,26 @@ def lc
   end
 
   describe ".order" do
-    it "supports virtual attributes symbol" do
+    it "orders by virtual attributes symbol" do
       Author.order(:nick_or_name).first # fails
```
NickLaMuro, line 820: Hey!  What the heck, man!  I definitely didn't authorize the use of my "name" in these tests!

spec/virtual_attributes_spec.rb
```diff
@@ -816,17 +816,26 @@ def lc
...
       Author.order(Author.arel_attribute(:nick_or_name)).first
     end
+
+    it "orders by virtual attributes inline arel" do
+      Author.order(Author.arel_attribute(:nick_or_name).asc).first
+    end
+
+    it "orders by virtual attributes inline arel" do
+      asc_node = Arel::Nodes::Descending.new(Author.arel_attribute(:nick_or_name))
+      Author.order(asc_node).first
```
NickLaMuro, line 837: So none of these tests have actual assertions?  That seems... not good.

NickLaMuro: I also see some of these with the comment of `# fails`... does that mean the specs don't pass?

NickLaMuro: FYI:  This don't have to be addressed in this PR, since that isn't the intent, just curious how these new tests should related to the other ones.  Obviously these would raise an error if they were invalid, but it probably also makes sense that they are at least returning data that we expect them to as well (again, not a requirement, just my 2cents)

spec/virtual_attributes_spec.rb
kbrock: nick or name looks like a typo for Nick **for** name

also reminds me of trick or treat

PR #77, opened 2020-08-23
Introduce aggregate methods for average, min, max

d-m-u:
blocked by:

- [x] https://github.com/ManageIQ/activerecord-virtual_attributes/pull/15 

adding `virtual_average`, `virtual_minimum`, `virtual_maximum` to the work started about a year ago that handles the nil/0 discrepancy where ActiveRecord/ruby treat the sum of no rows as 0 while SQL treats it as null

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_total.rb
+++ b/lib/active_record/virtual_attributes/virtual_total.rb
@@ -5,7 +5,7 @@ module VirtualAttributes
     module ClassMethods
       private
 
-      # define an attribute to calculating the total of a has many relationship
+      # define an attribute to calculate the total of a has many relationship
       #
       #  example:
       #
@@ -25,17 +25,38 @@ module VirtualAttributes
       #   # arel == (SELECT COUNT(*) FROM vms where ems.id = vms.ems_id)
       #
       def virtual_total(name, relation, options = {})
-        define_virtual_size_method(name, relation)
         define_virtual_aggregate_attribute(name, relation, :count, Arel.star, options)
+        define_method(name) { (has_attribute?(name) ? self[name] : send(relation).try(:size)) || 0 }
       end
 
-      # define an attribute to calculate the sum of a has may relationship
+      def virtual_sum(name, relation, column, options = {})
+        define_virtual_aggregate_attribute(name, relation, :sum, column, options)
+        define_virtual_aggregate_method(name, relation, column, :sum)
+      end
+
+      def virtual_minimum(name, relation, column, options = {})
+        define_virtual_aggregate_attribute(name, relation, :minimum, column, options)
+        define_virtual_aggregate_method(name, relation, column, :min, :minimum)
+      end
+
+      def virtual_maximum(name, relation, column, options = {})
+        define_virtual_aggregate_attribute(name, relation, :maximum, column, options)
+        define_virtual_aggregate_method(name, relation, column, :max, :maximum)
+      end
+
+      def virtual_average(name, relation, column, options = {})
+        define_virtual_aggregate_attribute(name, relation, :average, column, options)
+        define_virtual_aggregate_method(name, relation, column, :average) { |values| values.count == 0 ? 0 : values.sum / values.count }
+      end
+
+      #  @param method_name
+      #    :count :average :minimum :maximum :sum
       #
       #  example:
       #
       #    class Hardware
       #      has_many :disks
-      #      virtual_aggregate :allocated_disk_storage, :disks, :sum, :size
+      #      virtual_sum :allocated_disk_storage, :disks, :size
       #    end
       #
       #    generates:
@@ -54,9 +75,7 @@ module VirtualAttributes
 
       def virtual_aggregate(name, relation, method_name = :sum, column = nil, options = {})
         return virtual_total(name, relation, options) if method_name == :size
-
-        define_virtual_aggregate_method(name, relation, method_name, column)
-        define_virtual_aggregate_attribute(name, relation, method_name, column, options)
+        return virtual_sum(name, relation, column, options) if method_name == :sum
       end
 
       def define_virtual_aggregate_attribute(name, relation, method_name, column, options)
@@ -77,20 +96,19 @@ module VirtualAttributes
         end
       end
 
-      def define_virtual_size_method(name, relation)
-        define_method(name) do
-          (has_attribute?(name) ? self[name] : send(relation).try(:size)) || 0
-        end
-      end
-
-      def define_virtual_aggregate_method(name, relation, method_name, column)
+      def define_virtual_aggregate_method(name, relation, column, ruby_method_name, arel_method_name = ruby_method_name)
         define_method(name) do
           if has_attribute?(name)
             self[name] || 0
           elsif (rel = send(relation)).loaded?
-            rel.map { |t| t.send(column) }.compact.send(method_name)
+            values = rel.map { |t| t.send(column) }.compact
+            if block_given?
+              yield values
+            else
+              values.blank? ? nil : values.send(ruby_method_name)
+            end
           else
-            rel.try(method_name, column) || 0
+            rel.try(arel_method_name, column) || 0
           end
         end
       end
--- a/spec/db/models.rb
+++ b/spec/db/models.rb
@@ -27,7 +27,10 @@ class Author < VirtualTotalTestBase
            :class_name => "Book", :foreign_key => "author_id"
 
   virtual_total :total_recently_published_books, :recently_published_books
-  virtual_aggregate :sum_recently_published_books_rating, :recently_published_books, :sum, :rating
+  virtual_average :average_recently_published_books_rating, :recently_published_books, :rating
+  virtual_minimum :minimum_recently_published_books_rating, :recently_published_books, :rating
+  virtual_maximum :maximum_recently_published_books_rating, :recently_published_books, :rating
+  virtual_sum :sum_recently_published_books_rating, :recently_published_books, :rating
   virtual_delegate :description, :to => :current_photo, :prefix => true
 
   # This is here to provide a virtual_total of a virtual_has_many that depends upon an array of associations.
--- a/spec/virtual_total_spec.rb
+++ b/spec/virtual_total_spec.rb
@@ -579,6 +579,12 @@ RSpec.describe VirtualAttributes::VirtualTotal do
           end.to_not make_database_queries
         end
       end
+
+      it "orders by values with a nil (having the nil (defaulted to 0) first" do
+        authors
+        query = Author.order(:sum_recently_published_books_rating)
+        expect(query.map(&:id)).to eq([author3, author4, author2, author].map(&:id))
+      end
     end
   end
 end
--- a/spec/virtual_total_spec.rb
+++ b/spec/virtual_total_spec.rb
@@ -435,43 +435,150 @@ RSpec.describe VirtualAttributes::VirtualTotal do
       let(:author3) { Author.create }
       let(:author4) { Author.create.tap { |a| a.create_books(1, :published => true) } }
 
-      # NOTE: rails converts the nil to a 0
-      it "calculates sum with one off query" do
-        authors
+      describe ":sum" do
+        # NOTE: rails converts the nil to a 0
+        it "calculates sum with one off query" do
+          authors
 
-        expect do
-          expect(authors.map(&:sum_recently_published_books_rating)).to eq([6, 5, 0, 0])
-        end.to make_database_queries(:count => 4)
+          expect do
+            expect(authors.map(&:sum_recently_published_books_rating)).to eq([6, 5, 0, 0])
+          end.to make_database_queries(:count => 4)
+        end
+
+        it "calculates sum from preloaded association" do
+          authors.each { |a| a.recently_published_books.load }
+
+          expect do
+            expect(authors.map(&:sum_recently_published_books_rating)).to eq([6, 5, nil, nil])
+          end.to_not make_database_queries
+        end
+
+        it "calculates sum from attribute" do
+          authors
+          query = Author.select(:id, :sum_recently_published_books_rating).order(:id).load
+          expect do
+            expect(query.map(&:sum_recently_published_books_rating)).to eq([6, 5, 0, 0])
+          end.to_not make_database_queries
+        end
+
+        it "calculates sum from attribute (and preloaded association)" do
+          authors
+          query = Author.includes(:recently_published_books).select(:id, :sum_recently_published_books_rating).order(:id).load
+          expect do
+            expect(query.map(&:sum_recently_published_books_rating)).to eq([6, 5, 0, 0])
+          end.to_not make_database_queries
+        end
+
+        it "orders by values with a nil (having the nil (defaulted to 0) first" do
+          authors
+          query = Author.order(:sum_recently_published_books_rating)
+          expect(query.map(&:id)).to eq([author3, author4, author2, author].map(&:id))
+        end
       end
 
-      it "calculates sum from preloaded association" do
-        authors.each { |a| a.recently_published_books.load }
+      describe ":average" do
+        # NOTE: rails converts the nil to a 0
+        it "calculates avg with one off query" do
+          authors
 
-        expect do
-          expect(authors.map(&:sum_recently_published_books_rating)).to eq([6, 5, 0, 0])
-        end.to_not make_database_queries
+          expect do
+            expect(authors.map(&:average_recently_published_books_rating)).to eq([3, 5, 0, 0])
+          end.to make_database_queries(:count => 4)
+        end
+
+        it "calculates avg from preloaded association" do
+          authors.each { |a| a.recently_published_books.load }
+
+          expect do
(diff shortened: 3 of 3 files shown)


lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -77,6 +88,14 @@ def define_virtual_aggregate_attribute(name, relation, method_name, column, opti
         end
       end
 
+      %w[average maximum minimum].each do |m_name|
+        define_method("define_virtual_#{m_name}_method") do |name, relation, column|
+          define_method(name) do
+            (has_attribute?(name) ? self[name] : send(relation).try(m_name.to_sym, column)) || 0
```
kbrock, line 94: I think we may want to handle the case where the relation is loaded but the field is not. see for the size method.

It is too bad the size method takes no parameters, otherwise we could collapse all 4

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -77,6 +88,14 @@ def define_virtual_aggregate_attribute(name, relation, method_name, column, opti
         end
       end
 
+      %w[average maximum minimum].each do |m_name|
+        define_method("define_virtual_#{m_name}_method") do |name, relation, column|
```
kbrock, line 92: fun: this define method and metadata makes me squeamish.

Fun because everything here is using metadata to define_method.

spec/virtual_total_spec.rb
```diff
@@ -435,37 +435,144 @@ def model_with_children(count)
...
+            expect(authors.map(&:sum_recently_published_books_rating)).to eq([6, 5, nil, nil])
+          end.to_not make_database_queries
+        end
+
+        it "calculates sum from attribute" do
+          authors
+          query = Author.select(:id, :sum_recently_published_books_rating).order(:id).load
+          expect do
+            expect(query.map(&:sum_recently_published_books_rating)).to eq([6, 5, 0, 0])
+          end.to make_database_queries(:count => 2)
```
kbrock, line 461: I expected this to be zero. Any idea why it is hitting the database here?

maybe it is a string symbol kind of thing when calling `has_attribute?`

spec/virtual_total_spec.rb
```diff
@@ -435,37 +435,144 @@ def model_with_children(count)
...
-        expect do
-          expect(query.map(&:sum_recently_published_books_rating)).to eq([6, 5, nil, 0])
-        end.to_not make_database_queries
+      describe ":minimum" do
+        # NOTE: rails converts the nil to a 0
+        it "calculates min with one off query" do
+          authors
+
+          expect do
+            expect(authors.map(&:lowest_recently_published_books_rating)).to eq([2, 5, 0, 0])
```
kbrock, line 549: this is our standard.

We need to find a way to get the others to return the same value as this.

kbrock:
ok, some of these build failures is because a few blocks executed 0 queries instead of many (yay)

also a few returned 0 instead of nil (yay again)
so there are just a few little nil's that need to be tracked down and exterminated

lib/active_record/virtual_attributes/virtual_total.rb
d-m-u: not marking as resolved b/c i haven't addressed this yet

spec/virtual_total_spec.rb
d-m-u: i think we're good on this comment, i think i did mess up the `orders by values with a nil (having the nil (defaulted to 0) first` tests though

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -51,9 +52,19 @@ def virtual_total(name, relation, options = {})
...
+        method = "virtual_#{m_name}"
+        define_method(method) do |name, relation, column, options = {}|
+          send("define_#{method}_method", name, relation, column)
+          send("define_virtual_aggregate_attribute", name, relation, m_name, column, options)
+        end
+      end
 
       def virtual_aggregate(name, relation, method_name = :sum, column = nil, options = {})
-        return define_virtual_total(name, relation, options) if method_name == :size
+        return virtual_total(name, relation, options) if method_name == :size
```
d-m-u, line 64: so if method_name is any of these new three, @kbrock, we will never run `define_virtual_aggregate_method(name, relation, method_name, column)` which is what's checking to see if the data's been preloaded, i think, right?

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -42,7 +43,7 @@ def virtual_total(name, relation, options = {})
       #
       #    def allocated_disk_storage
       #      if disks.loaded?
-      #        disks.blank? ? nil : disks.map(&:size).compact.sum
+      #        disks.map(&:size).compact.sum
```
kbrock, line 46: don't think we're ready to make this declaration. yes, soon we will per another PR, but not yet.

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -51,9 +52,19 @@ def virtual_total(name, relation, options = {})
...
+      end
 
       def virtual_aggregate(name, relation, method_name = :sum, column = nil, options = {})
-        return define_virtual_total(name, relation, options) if method_name == :size
+        return virtual_total(name, relation, options) if method_name == :size
+        return virtual_average(name, relation, column, options) if method_name == :average
+        return virtual_minimum(name, relation, column, options) if method_name == :minimum
+        return virtual_maximum(name, relation, column, options) if method_name == :maximum
 
         define_virtual_aggregate_method(name, relation, method_name, column)
```
kbrock, line 69: is `sum` the only one left?

maybe do a case statement

lib/active_record/virtual_attributes/virtual_total.rb
kbrock: hmm, yea, lets sit on these for now but I'm definitely leaning towards skipping these and dumping this interface

lib/active_record/virtual_attributes/virtual_total.rb
kbrock: ok, if these are all the same, then no reason for duplicating the private method here.

Just call the contents of this directly from virtual_average or something

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -77,6 +90,14 @@ def define_virtual_aggregate_attribute(name, relation, method_name, column, opti
...
+        define_method("define_virtual_#{m_name}_method") do |name, relation, column|
+          define_method(name) do
+            (has_attribute?(name) ? self[name] : send(relation).try(m_name.to_sym, column)) || 0
+          end
+        end
+      end
+
       def define_virtual_size_method(name, relation)
         define_method(name) do
           (has_attribute?(name) ? self[name] : send(relation).try(:size)) || 0
```
kbrock, line 103: this works because relation (that is loaded).size is cached locally

yea, so this is here when the relation returns a nil - which is totally invalid.
Probably in the case where the relation is a virtual aggregate to a parent manager and the parent manager is not populated.

Should totally find another way to solve that problem

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -51,9 +52,19 @@ def virtual_total(name, relation, options = {})
...
       #    # arel => (SELECT sum("disks"."size") where "hardware"."id" = "disks"."hardware_id")
+      %w[average maximum minimum].each do |m_name|
+        method = "virtual_#{m_name}"
+        define_method(method) do |name, relation, column, options = {}|
+          send("define_#{method}_method", name, relation, column)
+          send("define_virtual_aggregate_attribute", name, relation, m_name, column, options)
+        end
+      end
 
       def virtual_aggregate(name, relation, method_name = :sum, column = nil, options = {})
```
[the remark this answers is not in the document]
kbrock: For the dsl, this method is not so good.

I'd prefer the user call the individual methods:

virtual_average
virtual_sum

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -85,10 +106,10 @@ def define_virtual_size_method(name, relation)
 
       def define_virtual_aggregate_method(name, relation, method_name, column)
         define_method(name) do
-          if attribute_present?(name)
-            self[name]
+          if has_attribute?(name)
+            self[name] || 0
           elsif (rel = send(relation)).loaded?
-            rel.blank? ? nil : rel.map { |t| t.send(column) }.compact.send(method_name)
+            rel.map { |t| t.send(column) }.compact.send(method_name)
```
kbrock, line 112: yea, this looks just like the changes in 15.

This method seems to handle all our use cases. No reason to define individual methods

lib/active_record/virtual_attributes/virtual_total.rb
kbrock: this is individual because the interface to determine this value is simply `relation.size` here

but for the others it is `relation.maximum(column)`

Average may be slightly different so we may want to handle that in its own method.
That or allow a block to be passed in

spec/db/models.rb
```diff
@@ -26,7 +26,13 @@ class Author < VirtualTotalTestBase
            :class_name => "Book", :foreign_key => "author_id"
 
   virtual_total :total_recently_published_books, :recently_published_books
+  virtual_average :average_recently_published_books, :recently_published_books, :rating
+  virtual_minimum :maximum_recently_published_books, :recently_published_books, :rating
+  virtual_maximum :minimum_recently_published_books, :recently_published_books, :rating
   virtual_aggregate :sum_recently_published_books_rating, :recently_published_books, :sum, :rating
+  virtual_aggregate :avg_recently_published_books_rating, :recently_published_books, :average, :rating
```
kbrock, line 33: yea, don't do these - stick with the pretty ones

spec/db/models.rb
```diff
@@ -26,7 +26,13 @@ class Author < VirtualTotalTestBase
            :class_name => "Book", :foreign_key => "author_id"
 
   virtual_total :total_recently_published_books, :recently_published_books
+  virtual_average :average_recently_published_books, :recently_published_books, :rating
+  virtual_minimum :maximum_recently_published_books, :recently_published_books, :rating
+  virtual_maximum :minimum_recently_published_books, :recently_published_books, :rating
   virtual_aggregate :sum_recently_published_books_rating, :recently_published_books, :sum, :rating
```
kbrock, line 32: maybe introduce `virtual_sum` (or pick a better name)

lets move away from the aggregate concept. At least for the public.

Rails has a calculate method. Users use `maximum` and `minimum` rather than `calculate(:maximum, ...)`
I think we should do the same

spec/virtual_total_spec.rb
```diff
@@ -435,37 +435,144 @@ def model_with_children(count)
...
+          end.to make_database_queries(:count => 4)
+        end
+
+        it "calculates min from preloaded association" do
+          authors.each { |a| a.recently_published_books.load }
```
kbrock, line 554: we have the concept of a relation, (`recently_published_books`) and a attribute (`lowest_recently_published_books_rating`) and an expected value (`[2, 5, 0, 0]`)
from there I think we can introduce an example.

I guess the relation is authors, and the sub relation is recently published book ratings, ... naming is hard

Just for this page for now.

lib/active_record/virtual_attributes/virtual_total.rb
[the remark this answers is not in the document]
NickLaMuro: > maybe we can deprecate this method.

I wouldn't be against the idea from what I can tell, but I would argue we can tackle in a separate PR.  I think that "how" we do that is enough of a discussion to be had where it probably makes sense to have it be a separate effort.

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -88,7 +91,13 @@ def define_virtual_aggregate_method(name, relation, method_name, column)
           if attribute_present?(name)
             self[name]
           elsif (rel = send(relation)).loaded?
-            rel.blank? ? nil : rel.map { |t| t.send(column) }.compact.send(method_name)
+            values = rel.map { |t| t.send(column) }.compact
+            return nil if values.blank?
+            if method_name == :average
```
kbrock: this is good.

I always had in my mind that we would pass a block into this method and `yield values` here but that is probably overkill

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -88,7 +91,13 @@ def define_virtual_aggregate_method(name, relation, method_name, column)
           if attribute_present?(name)
             self[name]
           elsif (rel = send(relation)).loaded?
-            rel.blank? ? nil : rel.map { |t| t.send(column) }.compact.send(method_name)
+            values = rel.map { |t| t.send(column) }.compact
+            return nil if values.blank?
```
kbrock:
```suggestion
            return nil if values.empty?
```

same thing but reads better to me. your choice

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -88,7 +91,13 @@ def define_virtual_aggregate_method(name, relation, method_name, column)
...
+            if method_name == :average
+              values.sum / values.size
```
kbrock: nice, zero won't come into here

spec/virtual_total_spec.rb
```diff
@@ -435,6 +435,7 @@ def model_with_children(count)
       let(:author3) { Author.create }
       let(:author4) { Author.create.tap { |a| a.create_books(1, :published => true) } }
 
+      describe ":sum" do
       # NOTE: rails converts the nil to a 0
       it "calculates sum with one off query" do
```
kbrock: probably want to indent these

kbrock (commented):
since rails does not implement average, we get to decide what we want to return. so we can defer to the sql behavior, which is to return a null

https://stackoverflow.com/questions/9661757/

spec/virtual_total_spec.rb
d-m-u: think you're looking at an old diff, sorry, these are totally indented currently

lib/active_record/virtual_attributes/virtual_total.rb
d-m-u: also think this one's an old diff, this is the current:
```ruby
if has_attribute?(name)
            self[name] || 0
          elsif (rel = send(relation)).loaded?
            values = rel.map { |t| t.send(column) }.compact
            if block_given?
              yield values
            else
              rel.blank? ? nil : values.send(ruby_method_name)
            end
          else
            rel.try(arel_method_name, column) || 0
          end
```

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -77,20 +96,19 @@ def define_virtual_aggregate_attribute(name, relation, method_name, column, opti
...
-          (has_attribute?(name) ? self[name] : send(relation).try(:size)) || 0
-        end
-      end
-
-      def define_virtual_aggregate_method(name, relation, method_name, column)
+      def define_virtual_aggregate_method(name, relation, column, ruby_method_name, arel_method_name = ruby_method_name)
         define_method(name) do
           if has_attribute?(name)
             self[name] || 0
           elsif (rel = send(relation)).loaded?
```
kbrock, line 103: huh, will this work on a virtual association that returns an array?
(sorry, scope creeping you)

you can use `virtual_has_many :books_with_authors` -- not the most optimized use case but should work

get the sum of an author virtual attribute for extra credit ;) -- though to be honest, that is probably not a good idea since we don't lazy load through virtual associations. Well, maybe we do but I'm not optimistic.

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -77,20 +96,19 @@ def define_virtual_aggregate_attribute(name, relation, method_name, column, opti
...
-            rel.map { |t| t.send(column) }.compact.send(method_name)
+            values = rel.map { |t| t.send(column) }.compact
+            if block_given?
+              yield values
+            else
+              rel.blank? ? nil : values.send(ruby_method_name)
```
kbrock, line 108: I think it should look at values maybe?
```suggestion
              values.blank? ? nil : values.send(ruby_method_name)
```

spec/db/models.rb
```diff
@@ -27,6 +27,9 @@ class Author < VirtualTotalTestBase
            :class_name => "Book", :foreign_key => "author_id"
 
   virtual_total :total_recently_published_books, :recently_published_books
+  virtual_average :average_recently_published_books_rating, :recently_published_books, :rating
+  virtual_minimum :minimum_recently_published_books_rating, :recently_published_books, :rating
+  virtual_maximum :maximum_recently_published_books_rating, :recently_published_books, :rating
   virtual_aggregate :sum_recently_published_books_rating, :recently_published_books, :sum, :rating
```
kbrock, line 33: since we are updating a major version for this, lets deprecate `virtual_aggregate` calls

kbrock (commented):
yea, just nits. lets merge this today

PR #78, opened 2020-08-23
fix typo in virtual aggregate guard clause

kbrock:
At this time, there is a different method virtual_total
that does count/size. this guard clause calls that other method for the user

We have changed all our code to use the proper `virtual_total` calls, hence we didn't get caught by this error. we are keeping this for safety reasons.

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_total.rb
+++ b/lib/active_record/virtual_attributes/virtual_total.rb
@@ -53,7 +53,7 @@ module VirtualAttributes
       #    # arel => (SELECT sum("disks"."size") where "hardware"."id" = "disks"."hardware_id")
 
       def virtual_aggregate(name, relation, method_name = :sum, column = nil, options = {})
-        return define_virtual_total(name, relation, options) if method_name == :size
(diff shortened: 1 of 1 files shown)


lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -53,7 +53,7 @@ def virtual_total(name, relation, options = {})
       #    # arel => (SELECT sum("disks"."size") where "hardware"."id" = "disks"."hardware_id")
 
       def virtual_aggregate(name, relation, method_name = :sum, column = nil, options = {})
-        return define_virtual_total(name, relation, options) if method_name == :size
+        return virtual_total(name, relation, options) if method_name == :size
```
jrafanie, line 56: What is the nature of this bug?  What is the difference between these two lines?

kbrock: there is no such thing as `define_virtual_total` - it is a syntax error
but since no one uses virtual aggregate with size, it has never showed up

so dropping this may also be a valid choice

PR #79, opened 2020-09-01
Fix delegates to start respecting polymorphic associations 

kbrock:
Before:

virtual delegates does not include the `type` column when fetching associated models.

So if a model has a polymorphic relationship (e.g. points to either books or authors), it would bring back extra records(e.g. books with author ids or authors with book ids).

It is a little bit of an edge case, and that is why this bug was not found until now. But since all tables start with an id of 1, the overlap can happen all the time.

In this case, you have to delegate from the has one side to the belongs to, which is uncommon for us and was not even possible before nick added the `limit()` and `order()` fixes.

In the bz example, the has_one only brings back one record, and it brought back a record with an id from a different class. So the limit (has one limits to one) brought back the wrong record.

After:

It properly constrains the records to the correct type.

Changes:

- test model has a new association
- test schema uses `references`
- test model has comments around use cases

I had looked into using `AssociationScope.scope` instead. It was very close, but is a much bigger change and would require a bit of monkey patching.

https://bugzilla.redhat.com/show_bug.cgi?id=1704644
https://github.com/ManageIQ/manageiq/issues/20115
https://github.com/ManageIQ/manageiq/pull/20494

/cc @NickLaMuro I'm changing the other va stuff as you read this

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_delegates.rb
+++ b/lib/active_record/virtual_attributes/virtual_delegates.rb
@@ -240,6 +240,7 @@ module ActiveRecord
       #     (SELECT "vms_sub"."name" FROM "vms" AS "vms_ss" WHERE "vms_ss"."id" = "vms"."src_template_id")
       #
 
+      # Based upon ActiveRecord AssociationScope.scope
       def self.select_from_alias(to_ref, col, to_model_col_name, src_model_id)
         query = if to_ref.scope
                   to_ref.klass.instance_exec(nil, &to_ref.scope)
@@ -247,12 +248,20 @@ module ActiveRecord
                   to_ref.klass.all
                 end
 
+        src_model   = to_ref.active_record
         to_table    = select_from_alias_table(to_ref.klass, src_model_id.relation)
         to_model_id = to_ref.klass.arel_attribute(to_model_col_name, to_table)
         to_column   = to_ref.klass.arel_attribute(col, to_table)
         arel        = query.except(:select).select(to_column).arel
                            .from(to_table)
                            .where(to_model_id.eq(src_model_id))
+
+        # :type is in the reflection definition (meaning it is polymorphic)
+        if to_ref.type
+          # get the class name (e.g. "Host")
+          polymorphic_type = src_model.base_class.name
+          arel = arel.where(to_ref.klass.arel_attribute(to_ref.type).eq(polymorphic_type))
+        end
 
         yield arel if block_given?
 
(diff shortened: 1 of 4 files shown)


lib/active_record/virtual_attributes/virtual_delegates.rb
```diff
@@ -247,13 +247,21 @@ def self.select_from_alias(to_ref, col, to_model_col_name, src_model_id)
...
 
+        src_model   = to_ref.active_record
         to_table    = select_from_alias_table(to_ref.klass, src_model_id.relation)
         to_model_id = to_ref.klass.arel_attribute(to_model_col_name, to_table)
         to_column   = to_ref.klass.arel_attribute(col, to_table)
         arel        = query.except(:select).select(to_column).arel
                            .from(to_table)
                            .where(to_model_id.eq(src_model_id))
 
+        # :type is in the reflection definition (meaning it is polymorphic)
```
kbrock, line 258: this is the only change

all other changes are either modifying the test schema (model) for demonstrating this issue

Our code roughly maps to [association_scope](https://github.com/rails/rails/blob/5-2-stable/activerecord/lib/active_record/associations/association_scope.rb#L21)

that method is focused on associations (that is to a particular base record) rather than a relation (that is general template for how records interact)

Now that we have added the `type` part, we have all the part implemented (though our `from().where()` construction is a little lacking)

spec/db/models.rb
```diff
@@ -12,6 +12,8 @@ class Author < VirtualTotalTestBase
   has_many :wip_books,       -> { wip },       :class_name => "Book"
   has_and_belongs_to_many :co_books,           :class_name => "Book"
   has_many :bookmarks,                         :class_name => "Bookmark", :through => :books
+  # who thinks I am targeting them. not who am I targeting
+  has_one :primary_target, -> { all.merge(Book.order(:name)) }, :as => :focus, :class_name => "Book"
```
Fryguy, line 16: would primary_book be clearer?

Fryguy: or maybe `focused_on`

spec/db/models.rb
```diff
@@ -94,6 +99,8 @@ class Book < VirtualTotalTestBase
   belongs_to :author
   has_and_belongs_to_many :co_authors, :class_name => "Author"
   belongs_to :author_or_bookmark, :polymorphic => true, :foreign_key => "author_id", :foreign_type => "author_type"
+  # person or book currently focusing on this book.
+  belongs_to :focus, :polymorphic => true
```
Fryguy, line 103: And then this can be `currently_focused_by`

spec/db/models.rb
kbrock: I do worry about `focused_on` looking like a date field, but it does read better.

kbrock: the use case for this bug is a definite stretch.

it is not who I am focused on, but rather who I think is focused on my
and it can't be primary book because it is a polymorphic

kbrock: thanks. I went back to this and decided to introduce a new model, 'Photo'. It is a classic polymorphic model and I feel makes the code easier to understand.

PR #80, opened 2020-09-25
update to 3.0.0

d-m-u:
we said in https://github.com/ManageIQ/activerecord-virtual_attributes/pull/15 that we were updating the major version

@kbrock do you want the changelog in this PR or can I make you do it for me?


kbrock:
if you are changing the version, the changelog goes with this

you can close if you don't want to take this on

d-m-u:
I'm definitely happy taking this on.

PR #81, opened 2020-10-10
drop support for AR sub five dot two

d-m-u:
since we're on 5.2, do we need to keep the 5.1 code?

... eventually...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,6 @@ env:
 - DB=pg
 - DB=sqlite3
 gemfile:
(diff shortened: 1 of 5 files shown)


Fryguy:
"We" is relative.  From a pure gem perspective, users of this gem might be on 5.1.  If we drop support for 5.1, we'll have to cut a major version (which I'm fine with since 5.1 is not longer supported by Rails).

d-m-u:
I don't know if there's any plan to cut this version and I'd like to close this.

PR #85, opened 2020-10-28
[WIP] version 4.0

d-m-u:
dropping 5.1 support requires a major version update 

see https://github.com/ManageIQ/activerecord-virtual_attributes/pull/81


d-m-u:
marking as work in progress because I think we may not quite be ready for this yet if we want to get rails attribute changes into this release as well.

d-m-u:
closing cause we've got other changes that we'd like to get into this release

PR #89, opened 2020-12-17
don't use send for include VirtualTotal

d-m-u:
does this need a `send`?


d-m-u:
Yeah, this won't work anyway, it was a good try though.

kbrock:
There was a change around making send public with some cases. It is ruby version specific and I can never remember the rules around this

PR #91, opened 2021-02-05
drop support for AR sub five dot two

d-m-u:
I thought there was other stuff we wanted to change here before we can merge this (pretty sure KB has some plans for what else he wants in this next major version change) but at the moment I'm at a bit of a loss as to specifically what that "other stuff" might entail.


Fryguy:
Looks good to me.  I think it might require a major version release for semver, but that's ok.

@kbrock Please review.

kbrock:
kicking build so I can merge

PR #93, opened 2021-02-06
[wip] arel_attribute is deprecated in 6.1

d-m-u:
`arel_table` is suggested instead of `arel_attribute`

`DEPRECATION WARNING: arel_attribute is deprecated and will be removed from Rails 6.2`

is for future, requires the prior merge of the other two open prs 

is work in progress because the line I have here for the config deprecation warning isn't correct:

the configurations deprecation warning: 
`ActiveRecord::Base.configurations` now uses `configs_for`: we're not supposed to "...call methods on the `configurations` hash directly..." (see https://github.com/rails/rails/pull/33637)
```
    def default_hash(env = default_env)
      default = find_db_config(env)
      default.configuration_hash if default
    end
    alias :[] :default_hash
    deprecate "[]": "Use configs_for", default_hash: "Use configs_for"
```

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_arel.rb
+++ b/lib/active_record/virtual_attributes/virtual_arel.rb
@@ -31,7 +31,7 @@ module ActiveRecord
               arel
             end
           else
-            super
+            arel_table[column_name]
           end
         end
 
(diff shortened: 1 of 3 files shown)


kbrock:
The issue is `arel_attribute` won't be called at all - so our api endpoint has been removed. Looks like we'll need to tap into the arel_table or something.

PR #95, opened 2021-04-16
add skeleton github actions tests

d-m-u:
I think you all might want to do this sometime, maybe, or something kind of close.


Fryguy:
@bdunne Can you take a look here, since you've played with GitHub Actions?

kbrock:
did try and get something like this up and running at https://github.com/kbrock/activerecord-hash_options
Still a few snags for mysql, but the rest is working ok.

I found that the time to run the tests were dwarfed by the time to setup the environment - so running a single action against the 3 databases made more sense - it was literally 3x faster than running a database per action.

For this gem, scheduling a job probably takes the longest, then setup is 40s, and the actual test run is 8s. Being a little creative may make sense here as well.

.github/workflows/run_test_suite.yml
```diff
@@ -0,0 +1,80 @@
...
+          --health-cmd="mysqladmin ping"
+          --health-interval=10s
+          --health-timeout=5s
+          --health-retries=3
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - ruby: 2.7
```
kbrock, line 38: for active record 5.2 we want ruby 2.5

```suggest
          - ruby: 2.5
```

.github/workflows/run_test_suite.yml
```diff
@@ -0,0 +1,80 @@
...
+            activerecord: 52
+          - ruby: 2.7
+            activerecord: 60
+          - ruby: 2.7
```
kbrock, line 42: suggested version is 3.0 but should support anything greater than 2.5 - so is still good

.github/workflows/run_test_suite.yml
```diff
@@ -0,0 +1,80 @@
...
+          --health-timeout=5s
+          --health-retries=3
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - ruby: 2.7
+            activerecord: 52
+          - ruby: 2.7
```
kbrock, line 40: suggested version is 2.6 but should support anything less than 3.0 - so is still good

.github/workflows/run_test_suite.yml
```diff
@@ -0,0 +1,80 @@
...
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+      - name: Install gems
+        run: |
+          bundle config path vendor/bundle
+          bundle install --jobs 4 --retry 3
+      - name: Setup mysql test database
+        run: |
+          sudo /etc/init.d/mysql start
+          mysql --port ${{ job.services.mysql.ports['3306'] }} -uroot -ppassword -e 'create database IF NOT EXISTS virtual_attributes_test;'
```
kbrock, line 73: password on the command line can be an issue here - think this is a FAQ to fix, but I don't have it.

.github/workflows/run_test_suite.yml
```diff
@@ -0,0 +1,80 @@
...
+      - name: Setup postgres test database
+        run: |
+          sudo /etc/init.d/postgresql start
+          sudo -u postgres psql -c "ALTER USER postgres PASSWORD 'postgres';"
+          psql -U postgres -c 'create database virtual_attributes_test;'
+      - name: Run pg tests
+        run: rake
```
kbrock, line 80: think we need to tell this test that we would like it to run PG or mysql / or sql lite

probably need to run for each of the databases

kbrock (commented):
Got my compatibility information from [fast ruby](https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html)

as a group we need to decide if we think running more test runs or grouping runs will make our lives easier to track down issues. more granular vs faster.

.github/workflows/run_test_suite.yml
d-m-u: thanks

kbrock:
thanks.

based another PR off of this
and then another PR off of that...

PR #97, opened 2022-01-08
Rails 5.2 changes (many needed for 6.1 support)

kbrock:
Lots of comment changes.
Some deprecations of rails 5.1 behavior

A few quick hits that are needed for rails 6.1 support.

commits: 

- stop using join_keys 
- comments to understand what changed in the cod
- more drop rails 5.1 changes
- drop deprecation warnings when associations are invalid 
- allow more recent versions of db query matchers (dev only)
- remove unneeded Arel::Attribute.to_sql 
- alphabetize gemspec
- remove send(:include)

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -115,7 +115,7 @@ module ActiveRecord
               super(reflection, records, scope)
             end
           end
-        elsif ActiveRecord.version.to_s >= "5.2" # < 6.0
+        else
           # preloader.rb active record 6.0
           # else block changed to reflect how 5.2 preloaders_for_one works
           def preloaders_for_reflection(reflection, records, scope, polymorphic_parent)
@@ -218,24 +218,6 @@ module ActiveRecord
             end
 
             ActiveSupport::Deprecation.warn(str, short_caller)
-          end
-        else
-          def preloaders_for_one(association, records, scope)
-            klass_map = records.compact.group_by(&:class)
-
-            # new logic: preload virtual fields / virtual includes
-            loaders = klass_map.keys.group_by { |klass| klass.virtual_includes(association) }.flat_map do |virtuals, klasses|
-              subset = klasses.flat_map { |klass| klass_map[klass] }
-              preload(subset, virtuals)
-            end
-            # /new logic
-
-            records_with_association = klass_map.select { |k, _rs| k.reflect_on_association(association) }.flat_map { |_k, rs| rs }
-            if records_with_association.any?
-              loaders.concat(super(association, records_with_association, scope))
-            end
-
-            loaders
           end
         end
       })
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -363,7 +363,10 @@ module ActiveRecord
       # From ActiveRecord::QueryMethods (rails 5.2 - 6.0)
       def build_select(arel)
         if select_values.any?
+          # change: pass allow_alias=true
+          # build_select builds a SELECT, so we want aliases on the columns. (this is the only place we do)
           arel.project(*arel_columns(select_values.uniq, true))
+          # /change
         elsif klass.ignored_columns.any?
           arel.project(*klass.column_names.map { |field| arel_attribute(field) })
         else
@@ -372,15 +375,21 @@ module ActiveRecord
       end
 
       # from ActiveRecord::QueryMethods (rails 5.2 - 6.0)
+      # @param allow_alias include AS alias for the generated SQL
+      #        value is true for SELECT and false for all others (e.g.: GROUP BY or COUNT)
       def arel_columns(columns, allow_alias = false)
         columns.flat_map do |field|
           case field
           when Symbol
+            # change: pass allow_alias
             arel_column(field.to_s, allow_alias) do |attr_name|
+            # /change
               connection.quote_table_name(attr_name)
             end
           when String
+            # change: pass allow_alias
             arel_column(field, allow_alias, &:itself)
+            # /change
           when Proc
             field.call
           else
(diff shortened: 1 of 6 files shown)


kbrock:
re Hakiri: I am not going to limit rails versions here. that is for the app not a gem to decide

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -363,7 +328,9 @@ def apply_join_dependency(*args, **kargs, &block)
       # From ActiveRecord::QueryMethods (rails 5.2 - 6.0)
       def build_select(arel)
         if select_values.any?
+          # change (only place allow_alias is true - the select clause)
```
jrafanie, line 331: I'm not sure what this means: `(only place allow_alias is true - the select clause)`... maybe say why there is a change here?  I can't figure it out at a glance.

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -376,11 +343,15 @@ def arel_columns(columns, allow_alias = false)
         columns.flat_map do |field|
           case field
           when Symbol
+            # change
```
jrafanie, line 346: Again, maybe just say why there is a change for future you/me/others.

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -376,11 +343,15 @@ def arel_columns(columns, allow_alias = false)
...
             arel_column(field.to_s, allow_alias) do |attr_name|
+            # /change
               connection.quote_table_name(attr_name)
             end
           when String
+            # change
```
jrafanie, line 352: Ditto.  Why?

lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -151,4 +151,4 @@ def virtual_aggregate_arel(reflection, method_name, column)
   end
 end
 
-ActiveRecord::Base.send(:include, VirtualAttributes::VirtualTotal)
+ActiveRecord::Base.include VirtualAttributes::VirtualTotal
```
jrafanie, line 154: Weird, why did we use send before?

spec/virtual_includes_spec.rb
```diff
@@ -200,6 +200,22 @@
...
+    #
+    # 1. bringing back authors for no reason - double work
+    # 2. joining to authors table not including author_id in the select clause
+    #
+    # old versions of rails ignore select() values when they have references.includes
+    # looks like rails 6.1 is now respecting the select, and thus breaks.
+    #
+    # wasted work and it is missing author_id which is needed for the join
+    #
+    # think this test is here because we do this bad code miq
```
jrafanie, line 218: 👍

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -190,10 +190,8 @@ def grouped_records(orig_association, records, polymorphic_parent)
 
               case association
               when Symbol, String
-                # 4/24/20 we want to revert #67 once we handle all these error cases in our codebase.
                 reflection = record.class._reflect_on_association(association)
-                display_virtual_attribute_deprecation("#{record.class.name}.#{association} does not exist") if !reflection && !polymorphic_parent
-                next if !reflection || !record.association(association).klass
+                next if polymorphic_parent && !reflection || !record.association(association).klass
```
jrafanie, line 194: Interesting.  So, now, we'd do `next` for a nil/false `polymorphic_parent`.  Previously, was this a bug?

lib/active_record/virtual_attributes/virtual_total.rb
kbrock: A pervious version of ruby made `include` private. Sorry, can't remember the exact version.
Now that we are only supporting more recent versions of ruby, we can use the standard `include`

jrafanie: Ah, ok, thanks!

kbrock:
### `polymorphic_parent` feature

#### Rails 6.0.x

Rails 6.0.x introduced the feature to blow up when `includes()` was passed an invalid value. Our team was frustrated with a few bugs in our code that were due to not properly checking the includes values.

#### Adding Rails 5.2 and 6.0 support

When I introduced support for rails 5.2 and 6.0 in #62 I used the 6.0.x implementation of `group_records` for both versions of rails. It was close enough and I did not have the energy to construct 2 monkeypatchs for two almost identical implementations of this complicated method. Besides, we wanted the protection of the 6.0 implementation.

#### It sounded like a good idea

Well, it turned out that we had more invalid `includes()` than we had originally anticipated, and we were not ready for this change. So we introduced the deprecation warning instead of blowing up in #67 

#### Today
Since over the past 2 years we have fixed any invalid `includes()`, this PR removes the deprecation warning and goes back to standard rails 6.0. Well, standard if you ignore the fact that we added the virtual includes.

kbrock:
push: added comments, cops, and fixed merge conflict

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -372,15 +340,21 @@ def build_select(arel)
       end
 
       # from ActiveRecord::QueryMethods (rails 5.2 - 6.0)
+      # @param allow_alias include AS alias for the generated SQL
+      #        valis is true for SELECT and false for all others (e.g.: GROUP BY or COUNT)
```
jrafanie, line 344: typo:  valis == value?

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -390,19 +364,24 @@ def arel_columns(columns, allow_alias = false)
...
```
jrafanie, line 368: Same... valis...

jrafanie:
> sorry didn't answer:
>> Previously, was this a bug?

I literally 🤣  when I got to this part.  Thanks for noticing you gave a useful history lesson and remembered you still didn't answer that.

> before 62, we were too eager to call next. Well, rails was too eager too
> before 67, we properly were calling the next - thanks to the rails 6.0 fix
> after 67, we rolled back the fix a little.
> This PR reverts the rollback, so we end up with logic closer to what resides in 6.0

😎 Ok, thanks!

lib/active_record/virtual_attributes/virtual_fields.rb
jrafanie: This was answered below in https://github.com/ManageIQ/activerecord-virtual_attributes/pull/97#issuecomment-1010195099

PR #98, opened 2022-01-10
Using actions instead of travis ci

kbrock:
This is cargo culted from ancestry.
But since I had spent a lot of time on building that implementation that I didn't feel too guilty.

If you note, all databases are running as a single job. reasons:
- the container setup times are much bigger than the runtime. Condensing them got all jobs running much quicker.
- failure in one database tended to be a failure in other databases. The exact database that caused the failure was not much of a concern
- At first, I had a gem file per database/rails version. Consolidating a gem per rails version and ignoring the database turned out to be a much better decision. So applying that logic to the build system seemed to also make sense.


chessbyte:
You probably also want to change the link to Travis in https://github.com/ManageIQ/activerecord-virtual_attributes/blob/master/README.md

Fryguy:
I can merge for now once it goes green, but I'll likely update to match the ones we have elsewhere, so that we have consistency (and also because eventually I want to have a centralized github workflow.

PR #100, opened 2022-01-10
Update GitHub actions for consistency

Fryguy:
- Fix tests
- Make the services, test names, etc consistent with other ManageIQ
  repos
- Fix CodeClimate

@kbrock Please review.


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,79 @@
...
+    strategy:
+      matrix:
+        ruby-version:
+        - '2.7'
+        activerecord:
+        - "52"
+        - "60"
+    services:
+      postgres:
+        image: postgres:13
```
Fryguy, line 17: I was going to switch this over to manageiq/postgresql:10 for consistency, but for some reason I was getting unhealthy containers.  For now, this is fine.

spec/db/database.yml
```diff
@@ -21,3 +21,4 @@ mysql2:
   <<: *common
   adapter: mysql2
   username: root
+  host: 127.0.0.1
```
Fryguy, line 24: The MySQL tests would not work without this because localhost tells the mysql client to use a local socket which won't work in github actions.

Fryguy:
@kbrock It looks like an add and a delete, because I renamed the file from ci.yml to ci.yaml for consistency.

PR #101, opened 2022-01-12
Rails6.1

kbrock:
depends:
- [x] #97 
- [x] #103

Changes:

- update gemfiles and interface changes
- transition to using `arel_table` from `arel_attribute`
- drop support for `select(:vcol1).includes(:vcol1).references(:vcol1)` (plan is to update our reports to avoid this)
- remove quite a few monkey patches
- require `virtual_attributes :arel` returns a `Arel::Node::Grouping` node (but auto adds them)
 
Note: this is still rails 5.2, 6.0, and 6.1 compatible


.github/workflows/ci.yaml
```diff
@@ -12,6 +12,7 @@ jobs:
         activerecord:
         - "52"
         - "60"
+        - "61"
```
Fryguy, line 15: Do we want to add 7.0 as a another entry in the matrix, but allowed to fail?  If so, the way I did that elsewhere was like this:

```yaml
      matrix:
        ruby-version:
        - '2.7'
        activerecord:
        - "52"
        - "60"
        - "61"
        allow-failure: false
        include:
        - activerecord: "70"
          allow-failure: true
```

Then down below (in each of the test sections)

```yaml
      - name: Run SQLite tests
        continue-on-error: ${{ matrix.allow-failure }}
        env:
          DB: sqlite3
          CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
```

kbrock: I was definitely leaning towards no here. We are quite far away from 7.0

Do you want me to put it in anyway?

Anything else you want before we push this out?

jrafanie: I see no value in testing 7.0 if we can't even get useful failures or the error/failures are so numerous, it's too hard to read. I'll leave that up to you to determine how close we are to supporting 7.0

kbrock:
WIP: looks like aggregate of an aggregate broke and possibly some others

Will add some tests to properly test these edge cases

kbrock:
rubocop: good news, the complex method of concern is no longer used for rails 6.1 and above.
This branch is still supporting 5.2, 6.0, and 6.1 - otherwise, we could delete these

unwip: aggregation of aggregation was not an issue.
I did add support for arel_table[alias] to work. Not perfect but seems to resolve a few issues where we are counting virtual attributes and the like.

.github/workflows/ci.yaml
kbrock: not close enough.

Fryguy: sounds good...no need for 70 then :)

Fryguy:
Minor, but I just realized why I asked my question in https://github.com/ManageIQ/activerecord-virtual_attributes/pull/101#discussion_r783138599, and that's because I saw the 7.0 appraisal file and in the appraisal list.  Should we remove that from this PR and make that part of a separate PR?

kbrock:
push: fixed cops. removed the 7.0 appraisal

lib/active_record/virtual_attributes/virtual_arel.rb
```diff
@@ -14,6 +14,30 @@ class Arel::Nodes::Grouping
     end
 
     module VirtualArel
+      class ATProxy < Arel::Table
```
Fryguy, line 17: Just for my knowledge is `ATProxy` a Rails thing (that you are trying to override from Rails) or is that a construct specifically for this gem.?

lib/active_record/virtual_attributes/virtual_arel.rb
```diff
@@ -14,6 +14,30 @@ class Arel::Nodes::Grouping
...
+        attr_accessor :klass
+
+        def [](name, table = self)
```
Fryguy, line 20: Also is this code "our" code or code copied from Rails and then modified somehow?  If the latter, it might be useful to tack in a comment linking to where it came from and describing what was changed.  Then, in the git history, having a copy of the original with our changes on top in a separate commit so we can see the diff (I think this was kind of done already, but I figured I'd call out that out anyway)

lib/active_record/virtual_attributes/virtual_arel.rb
```diff
@@ -22,17 +46,13 @@ module VirtualArel
...
-              arel = col.call(arel_table)
-              arel.name = column_name if arel.kind_of?(Arel::Nodes::Grouping)
-              arel
-            end
-          else
-            super
-          end
+        # ActiveRecord::Core
+        def arel_table # :nodoc:
+          @arel_table ||= if ActiveRecord.version.to_s < "6.1"
```
Fryguy, line 51: Super minor, but can we extract the version check to outside of the method, maybe into a constant?  Otherwise we are constant looking up the version and converting to a string.

lib/active_record/virtual_attributes/virtual_delegates.rb
```diff
@@ -243,25 +247,30 @@ def self.select_from_alias(to_ref, col, to_model_col_name, src_model_id)
...
-        arel        = query.except(:select).select(to_column).arel
-                           .from(to_table)
-                           .where(to_model_id.eq(src_model_id))
+        src_model = to_ref.active_record
+        to_table  = select_from_alias_table(to_ref.klass, src_model_id.relation)
+        if ActiveRecord.version.to_s < "6.1"
+          to_model_id = to_ref.klass.arel_attribute(to_model_col_name, to_table)
+          to_column   = to_ref.klass.arel_attribute(col, to_table)
+        else
+          to_model_id = to_ref.klass.arel_table[to_model_col_name, to_table]
```
Fryguy, line 256: Is arel_table available on < 6.1?  If so, can we just switch to arel_table everywhere like how it was done in https://github.com/ManageIQ/manageiq-api/pull/1121 ?

lib/active_record/virtual_attributes/virtual_arel.rb
kbrock: it is specific for this Gem. ArelTableProxy - suggest a different name if you want

lib/active_record/virtual_attributes/virtual_arel.rb
kbrock: this is ours - actually copied from our original `arel_attribute` code
I'll add a comment defining what it does

lib/active_record/virtual_attributes/virtual_delegates.rb
kbrock: in 6.1, we want to use arel_table everywhere. there is a deprecation warning if we do not do this.

In 6.0, for the single attribute version, we want to use arel_table. But in the double parameter version, we still want to use `arel_attribute()`

I'll look into possibly providing `arel_attribute` type functionality to `arel_table`.

Since we are kinda dropping 6.0 support, didn't put that much effort into it

lib/active_record/virtual_attributes/virtual_arel.rb
kbrock: This will be duplicate code.
I agree with the premise. Just hope that there are not nuances that I missed with these changes

kbrock:
UPDATE: pretty big rework.

- all version checks have been extracted outside methods
- extracted virtual arel lookup outside rails methods in another commit.
- removed cops around complexity (and deeply nested `if` complaint)
- comments around purpose of more methods
- still ignoring instantiate since that goes away in 6.1 and above

kbrock:
update:

- dropped code for 5.2
- using YAML.safe_load (denormalized database.yml)
- auto add a grouping (think parens) node when it is missing

kbrock:
update:
- extracted grouping into PR 103 and rebased to beginning of changes.

lib/active_record/virtual_attributes/virtual_arel.rb
kbrock: extracted to another method. duplication be gone

lib/active_record/virtual_attributes/virtual_arel.rb
kbrock: added comments to explain what is happening. hope this fixes the issues

lib/active_record/virtual_attributes/virtual_arel.rb
kbrock: changed the name and added a comment.

spec/support/database.rb
```diff
@@ -27,8 +27,13 @@ def setup
 
   def migrate
     ActiveRecord::Migration.verbose = false
-    ActiveRecord::Base.configurations = YAML.safe_load(ERB.new(IO.read("#{dirname}/database.yml")).result)
-    ActiveRecord::Base.establish_connection ActiveRecord::Base.configurations[adapter]
+    databaseyml = YAML.safe_load(ERB.new(IO.read("#{dirname}/database.yml")).result)
+    if ActiveRecord::VERSION::STRING >= "6.1"
```
jrafanie, line 31: Does it make sense to flip this conditional or flip the others?  I think most of the others do `if < 6.1`.  I'd be fine with having this format above and change the others so it's obvious that 6.1+ is the expected outcome going forward and have the `else` handle the less expected cases.

jrafanie (commented):
I don't follow all of the changes but as long as cross repo on this works with both 6.0 and 6.1, I think we can move forward.  I had a few comments/questions above.

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -327,80 +327,29 @@ def apply_join_dependency(*args, **kargs, &block)
...
 
-      # From ActiveRecord::QueryMethods (rails 5.2 - 6.0)
+      # From ActiveRecord::QueryMethods (rails 5.2 - 6.1)
       def build_select(arel)
         if select_values.any?
-          # change: pass allow_alias=true
-          # build_select builds a SELECT, so we want aliases on the columns. (this is the only place we do)
-          arel.project(*arel_columns(select_values.uniq, true))
-          # /change
-        elsif klass.ignored_columns.any?
```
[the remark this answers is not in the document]
kbrock: ### before

copy rails build_select and add our changes

### after

use `super` (which implements this case and the others) for all but the case where the user gives us `select()` values (and in that case, lets swap out virtual attributes)

kbrock:
update:

- changed version check to be `< "6.1"

LJ's question is being handled by active record proper

kbrock:
ok. this should be close.
Any final comments?

this will go into virtual attributes version 6.0.
We will then drop rails version 6.0 support and put into virtual attributes version 6.1

jrafanie (approved):

Fryguy:
As discussed let's:

- Change the changelog line for 6.0 to 6.1:

   ```diff
   - ## [6.0.0] - 2022-02-02
   + ## [6.1.0] - 2022-02-02
   ```

- Drop the changelog line that mentions 6.0

   ```diff
   - e.g.: virtual attributes 6.0.x supports all versions of rails 6.0.
   -    virtual attributes 6.1.x supports all versions of rails 6.1.
   + e.g. virtual attributes 6.1.x supports all versions of rails 6.1.
   ```

- Split out dropping 5.2 into a separate PR
- Create a separate PR for dropping 6.0

kbrock:
push:

- did not drop 5.2
- changed changelog

jrafanie:
LGTM

PR #103, opened 2022-01-26
Auto add grouping

kbrock:
### Background

There has always been an non-enforced rule that the arel generated for a virtual_attribute needs to be wrapped in parenthesis. In Arel, this is called "grouping" aka `Arel::Nodes::Grouping`

### Before

If the Arel returned from `virtual_attribute :arel => lambda {}` is not a `Grouping`, the virtual attribute will be silently ignored.

### After

If the Arel returned from `virtual_attribute :arel => lambda {}` is not a `Grouping`, the node will be converted to a `Grouping`


jrafanie (approved):
LGTM

PR #105, opened 2022-02-02
Drop rails 5.2 support

kbrock:
we are dropping active record 5.2 for our virtual attributes 6.1 release

PR #106, opened 2022-02-03
Drop rails 6.0 support

kbrock:
DEPENDS on:

- https://github.com/ManageIQ/activerecord-virtual_attributes/pull/105

Only 1 commit in this PR.


.github/workflows/ci.yaml
```diff
@@ -9,9 +9,6 @@ jobs:
       matrix:
         ruby-version:
         - '2.7'
-        activerecord:
-        - "60"
-        - "61"
```
chessbyte, line 14: don't you want to keep 6.1?

jrafanie: Or wait, it is just doing the default of rails 6.1 from the Gemfile?

chessbyte: but, at some point we will want to validate Rails 7.0 - so I assume we want to be able to specify multiple versions of ActiveRecord?

kbrock: we will be doing a different branch for each version

when we are ready for 7.0, then we change across.
if we need to backport to 6.1.1, we will test those changes for 6.1 only

does this plan still sound good?

chessbyte: sure - just wanted to ensure it was thought through!

Fryguy: @kbrock is correct...since we are moving to a "branch/release per Rails version", there's no longer a need to support multiple versions simultaneously.  The 6-1-stable branch or whatever we call it will be 6.1 only and master will be 7.0 until we cut a 7-0-stable branch.

.github/workflows/ci.yaml
```diff
@@ -47,7 +44,6 @@ jobs:
       MYSQL_HOST: 127.0.0.1
       MYSQL_PWD: password
       # for rails tests (from matrix)
```
Fryguy, line 46: This comment can go away too.

Fryguy (approved):

kbrock:
update:
- removed unnecessary matrix comment

jrafanie (approved):
LGTM

PR #107, opened 2022-02-03
Move versioning verbiage into README

Fryguy:
[skip ci]

This makes it more clear up front, before installing, what version to pick.  The changelog now references the README for the details.


kbrock:
thanks. this reads better

PR #111, opened 2022-07-22
Fix virtual attributes and IN clauses

kbrock:
6.1 introduced `HomogeneousIn`, an optimization for the `In` clause.
This introduced changes to `Attribute` that required our attention.

An attribute now needs to have a `name`, and be able to type cast parameters for the
datatype.

This change gives the virtual attribute more information for type casting.

Put in a PR for rails to hopefully remove our requirement to monkey patch
that operator. https://github.com/rails/rails/pull/45642

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_arel.rb
+++ b/lib/active_record/virtual_attributes/virtual_arel.rb
@@ -98,3 +98,43 @@ module ActiveRecord
     end
   end
 end
+
+module Arel # :nodoc: all
+  # rubocop:disable Naming/MethodName
+  # rubocop:disable Naming/MethodParameterName
+  # rubocop:disable Style/ConditionalAssignment
+  module Visitors
+    # rails 6.1...
+    class ToSql
+      private
+
+      def visit_Arel_Nodes_HomogeneousIn(o, collector)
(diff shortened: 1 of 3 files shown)


Fryguy:
> Put in a PR for rails to hopefully remove our requirement to monkey patch
that operator.

Can you link the Rails PR?

jrafanie:
To be clear, does this PR work with rails 6.0?  I'm good with this change but am curious how we'll version this if it no longer works with 6.0.

kbrock:
@jrafanie Our versions are for a single version of rails.

This change is only for HEAD / virtual attributes 6.1.x / ActiveRecord 6.1.x
If we want a change for 6.0, then we can backport to an older version.

So yes, this will not break previous versions of rails because they will not use this version of virtual attributes.

jrafanie:
> @jrafanie Our versions are for a single version of rails.
> 
> This change is only for HEAD / virtual attributes 6.1.x / ActiveRecord 6.1.x If we want a change for 6.0, then we can backport to an older version.

Right.  Thanks.

jrafanie (approved):

jrafanie:
We'll need a tagged release for this change.

PR #114, opened 2022-08-09
Introduce Arel::Nodes::VirtualAttribute

kbrock:
## Overview

Virtual attributes need an attribute name. The node representing the attribute needs to return a name and type.

In order to work with `ToSql`, this needs to be in the `Arel::Nodes::` namespace.

## Before

- `Arel::Nodes::Grouping` was monkeypatched to return `name` and `relation`.

Reminder: This node represents the parenthesized grouping of sql.

## After

- Remove monkey patch from `Arel::Nodes::Grouping`
- Introduce `Arel::Nodes::VirtualAttribute` that is basically a grouping
- `ToSql` treats this node as a `Grouping` (since that is the parent class)

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_arel.rb
+++ b/lib/active_record/virtual_attributes/virtual_arel.rb
@@ -10,8 +10,13 @@ module ActiveRecord
     # Model.select(Model.arel_table.grouping(Model.arel_table[:field2]).as(:field))
     # Model.attribute_supported_by_sql?(:field) # => true
 
-    # in essence, this is our Arel::Nodes::VirtualAttribute
-    class Arel::Nodes::Grouping
+    class VirtualAttribute < Arel::Nodes::Grouping
+      def initialize(arel, name = nil, relation = nil)
+        super(arel)
+        @name = name
+        @relation = relation
+      end
+
       attr_accessor :name, :relation
 
       # methods from Arel::Nodes::Attribute
@@ -95,10 +100,12 @@ module ActiveRecord
           return unless arel_lambda
 
           arel = arel_lambda.call(table)
-          arel = Arel::Nodes::Grouping.new(arel) unless arel.kind_of?(Arel::Nodes::Grouping)
-          arel.name = column_name
-          arel.relation = table
-          arel
+          # By convention, all attributes are defined with a grouping.
+          # Since we're adding a VirtualAttribute node, which is essentially a
+          # grouping, there is no need to keep both and end up with double parens
+          arel = arel.expr if arel.kind_of?(Arel::Nodes::Grouping)
+
+          VirtualAttribute.new(arel, column_name, table)
         end
 
         private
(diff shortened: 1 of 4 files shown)


kbrock:
sigh

def or(other)
  left = self - other
  common = self - left
  right = other - common

  if left.empty? || right.empty?
    common
  else
    left = left.ast
    left = left.expr if left.is_a?(Arel::Nodes::Grouping)

    right = right.ast
    right = right.expr if right.is_a?(Arel::Nodes::Grouping)

    or_clause = Arel::Nodes::Or.new(left, right)

    common.predicates << Arel::Nodes::Grouping.new(or_clause)
    common
  end
end

this is why we can't have nice things.
Wonder why they didn't use kind_of?(Grouping)

Introducing our own node type may be troublesome
But I setup tests with In and Or (that are polymorphic) and the seem to be good

kbrock:
I think when I went back to my comment, I realized that having the `Or` the way it is actually bought us some benefits. So no longer bummed about that.

This PR is more of a "what if" than a "we should do this".

It is based on the notion that it would be nice if we used `ToSql` to generate this sql rather than hacking the callers

kbrock:
Removing the modifications to `Group` and having our own node (that extends and works like a `Group`) will give us some autonomy

I will add this to my existing branch that I'm running with cross repo

kbrock:
un-WIP: I had WIP'd this before because I was concerned that some unintended consequence could sneak through. Now that we have to run more thorough tests to get rails 7.0 working, it seems a good time to move forward with this.

The only possible code change outstanding is to update `ToSql` and remove some of our relation monkey patches. But these were not trivial and moving forward with this change is currently less risky.

lib/active_record/virtual_attributes/virtual_arel.rb
```diff
@@ -11,7 +11,13 @@ module VirtualAttributes
     # Model.attribute_supported_by_sql?(:field) # => true
 
     # in essence, this is our Arel::Nodes::VirtualAttribute
-    class Arel::Nodes::Grouping
+    class ::Arel::Nodes::VirtualAttribute < Arel::Nodes::Grouping
```
kbrock, line 14: It would have been nice if our attribute was in our namespace, but Arel requires that all nodes be in their namespace

kbrock:
update:
- rebase (resolve conflict)

lib/active_record/virtual_attributes/virtual_arel.rb
jrafanie: hmm, that's too bad.  Do you think they don't want this exposed?  I wonder if there's another way.  Can we use composition with a module providing the behavior instead?  I'd be afraid of adding to their namespace.  I guess we'd still be including a module in their namespace.  Maybe, it's moot then.

kbrock: YAY - thank you. This works in our own namespace.

fixed

kbrock:
update:
- moved VirtualAttribute out of Arel and into to our namespace

Maybe it is working because `ToSql` is handling `Grouping` so it is fine.
We may need to move if we add a custom handler in `ToSql`, but don't think that will be necessary any time soon (famous last words)

jrafanie (approved):
:shipit:

PR #116, opened 2022-08-15
spec: ensure delegates work with OR and HomogeneousIn

kbrock:
extracted from #114 

mostly focused on ensuring the arel coming from the delegate works correctly.

I thought we had the grouping and to_sql in VirtualDelegate#select_from_alias
was to ensure that binding variables did not get botched,
but I can't reproduce the issue and get it to fail.

Adding these specs for now and hoping to get a good reproducer for the need
of the `Arel.sql(arel.to_sql)`

### DIFF
--- a/spec/virtual_delegates_spec.rb
+++ b/spec/virtual_delegates_spec.rb
@@ -236,14 +236,48 @@ RSpec.describe ActiveRecord::VirtualAttributes::VirtualDelegates, :with_test_cla
   end
 
   context "with polymorphic has_one" do
-    it "respects type" do
+    it "supports select" do
       author = Author.create(:name => "no one of consequence")
-      book = author.books.create(:name => "nothing of consequence", :id => author.id)
       author.photos.create(:description => 'good')
-      book.photos.create(:description => 'bad')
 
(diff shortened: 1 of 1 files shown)


kbrock:
Rails did just merge a fix for `HomogeniusIn`. So it will maybe be in 7.0.5, but more likely 7.1

@jrafanie this is only tests, you ok merging this?

kbrock:
update:
- rebased

spec/virtual_delegates_spec.rb
```diff
@@ -236,14 +236,50 @@ def self.connection
   end
 
   context "with polymorphic has_one" do
+    it "supports select" do
+      Author.delete_all
+      Book.delete_all
```
jrafanie, line 241: I don't understand the need to delete authors and books in these tests. I don't see other tests creating them.  Do we have some setup that creates these and we're trying to remove them before we create the objects we want to test?

jrafanie (approved):
LGTM, just have a minor question about the common cleanup before the tests

jrafanie (approved):
Let's merge as is.  We can determine if there's a better way to ensure data is rolled back so we don't need to cleanup before each test.

PR #117, opened 2023-04-28
add ruby 3.1 to test matrix

kbrock:
I noticed that we were still testing 2.7
Adding 3.1 support

Nothing happened to make me think this will break, but wanted to put putting this in place.

remember: We are linking our version to active record version numbers, so this branch (master is rails 6.1.x specific


jrafanie (approved):
LGTM

PR #118, opened 2023-04-30
Use database_cleaner to clear out tables

kbrock:
We would like to clean up all records created in each test.
This was not properly working before so we had to manually delete records ourselves.

Rails ActiveRecord does provide `transactional_features` for tests, but it is not
so simple to implement. This functionality is provided by `rspec-rails`, but not
in standard `rspec` (which we are using).

Using DatabaseCleaner gem to clean up after each test.
The `transaction` strategy should work for us since we only use a single database
and a single connection.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -8,6 +8,7 @@ $LOAD_PATH.unshift(File.dirname(__FILE__) + '/../lib')
 require "bundler/setup"
 require "active_record/virtual_attributes"
 require "active_record/virtual_attributes/rspec"
+require "database_cleaner/active_record"
 require "db-query-matchers"
 
 Dir['./spec/support/**/*.rb'].sort.each { |f| require f }
@@ -22,5 +23,18 @@ RSpec.configure do |config|
 
   config.expect_with :rspec do |c|
     c.syntax = :expect
+  end
+
+  config.before(:suite) do
+    # truncate at startup
+    DatabaseCleaner.clean_with :truncation
+    # transaction between examples (mysql requires truncation)
+    DatabaseCleaner.strategy = ENV["DB"].to_s.include?("mysql") ? :truncation : :transaction
+  end
+
+  config.around(:each) do |example|
+    DatabaseCleaner.cleaning do
+      example.run
+    end
   end
 end
--- a/spec/virtual_total_spec.rb
+++ b/spec/virtual_total_spec.rb
@@ -1,10 +1,4 @@
 RSpec.describe VirtualAttributes::VirtualTotal do
-  before do
-    Author.delete_all
-    Book.delete_all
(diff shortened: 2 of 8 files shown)


kbrock:
Sorry, this is becoming a catch-all. Since they are all test related, grouping together - though if you want them as separate PRs, let me know.

update:
- added conditional mysql logic (mysql is requiring truncate for some reason)
- tweaked slowest test to run quicker
update:
- fixed cops around conditional logic. regexp => `String.include?()`
- added bin/setup to create the database

spec/spec_helper.rb
```diff
@@ -23,4 +24,16 @@
   config.expect_with :rspec do |c|
     c.syntax = :expect
   end
+
+  config.before(:suite) do
+    # mysql requires truncation
+    DatabaseCleaner.strategy = ENV["DB"].to_s.include?("mysql") ? :truncation : :transaction
+    DatabaseCleaner.clean_with(:truncation)
```
jrafanie, line 31: Is this correct?  We specify the strategy of `transaction` on line 30 for pg but then `clean_with` is passed `truncation`?

spec/virtual_total_spec.rb
```diff
@@ -1,10 +1,4 @@
 RSpec.describe VirtualAttributes::VirtualTotal do
-  before do
-    Author.delete_all
-    Book.delete_all
-    Bookmark.delete_all
-  end
```
jrafanie, line 6: Yay, all the red

spec/spec_helper.rb
kbrock: yea. so I think they have 2 concepts.
What you should do at the beginning of everything (the `suite`), and what you should do after each transaction.

I did this based upon the bottom of their [how to use](https://github.com/DatabaseCleaner/database_cleaner#how-to-use) section

jrafanie: > I did this based upon the bottom of their [how to use](https://github.com/DatabaseCleaner/database_cleaner#how-to-use) section

Thanks, I get it now. Use this strategy going forward but first truncate the db.

Should we swap the order like they have in that section?

image

spec/spec_helper.rb
```diff
@@ -23,4 +24,17 @@
   config.expect_with :rspec do |c|
     c.syntax = :expect
   end
+
+  config.before(:suite) do
+    # truncate at startup
+    DatabaseCleaner.clean_with :truncation
+    # transaction between examples (mysql requires truncation)
+    DatabaseCleaner.strategy = ENV["DB"].to_s.include?("mysql") ? :truncation : :transaction
```
jrafanie, line 32: 🥇 Thanks, this helps a lot!

jrafanie (approved):

PR #120, opened 2023-06-01
Handle 'uses' delegating to another 'uses' that has a deep hash

kbrock:
## Issue

The use case that triggered this had more layers of indirection.
But the jist is when we map a `:uses` with an array, and one of the entries has a complex `:uses`,
it will end up with a nested hash.

```
virtual_has_many :field1, :uses => {:a => :b}
virtual_has_many :field2, :uses => :c
virtual_has_many :field3, :uses => [:field1, :field2]
```

## Before

```
references([:field3])
=>
references([:field1, :field2])
=>
references([{:a => :b}, :c]
```

And when we map that into a hash we end up with
```
{{:a => :b} => {}, :c => {}}
```
where `{:a => :b}` is a key pointing to an empty hash (i.e.: `{}`)

### So we blow up

```
Issue with `Vm#used_storage`:

ERROR -- evm: [NoMethodError]: undefined method `to_sym' for {:hardware=>:disks}:Hash
Did you mean?  to_s
               to_set
ERROR -- evm: lib/active_record/virtual_attributes/virtual_reflections.rb:36:in `virtual_reflection?`
lib/active_record/virtual_attributes/virtual_fields.rb:30:in `virtual_field?`
lib/active_record/virtual_attributes/virtual_fields.rb:50:in `block in replace_virtual_field_hash`
lib/active_record/virtual_attributes/virtual_fields.rb:49:in `each`
```

## After

We don't assume that all entries in the input array are a string/symbol:
so we end up with simple keys in the destination hash:

```
{:a => :b, :c => {}}
```

Fixes https://github.com/ManageIQ/manageiq/issues/22526

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -69,7 +69,9 @@ module ActiveRecord
           when String, Symbol
             {value => {}}
           when Array
-            value.flatten.each_with_object({}) { |k, h| h[k] = {} }
+            value.flatten.each_with_object({}) do |k, h|
+              merge_includes(h, k)
+            end
           when nil
             {}
           else
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -77,15 +77,14 @@ module ActiveRecord
           end
         end
 
-        # @param [Hash] hash1
-        # @param [Hash] hash2
+        # @param [Hash] hash1 (incoming hash is modified and returned)
+        # @param [Hash|Symbol|nil] hash2 (this hash will not be modified)
         def merge_includes(hash1, hash2)
           return hash1 if hash2.blank?
 
-          hash1 = include_to_hash(hash1)
-          hash2 = include_to_hash(hash2)
-          hash1.deep_merge!(hash2) do |_k, v1, v2|
-            merge_includes(v1, v2)
+          hash1.deep_merge!(include_to_hash(hash2)) do |_k, v1, v2|
+            # this block is conflict resolution when a key has 2 values
+            merge_includes(include_to_hash(v1), v2)
           end
         end
       end
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -84,6 +84,13 @@ module ActiveRecord
         def merge_includes(hash1, hash2)
           return hash1 if hash2.blank?
 
+          # very common case.
+          # optimization to skip deep_merge and hash creation
+          if hash2.kind_of?(Symbol)
+            hash1[hash2] ||= {}
+            return hash1
+          end
+
           hash1.deep_merge!(include_to_hash(hash2)) do |_k, v1, v2|
             # this block is conflict resolution when a key has 2 values
             merge_includes(include_to_hash(v1), v2)
--- a/spec/db/models.rb
+++ b/spec/db/models.rb
@@ -126,6 +126,8 @@ class Book < VirtualTotalTestBase
   # this tests delegate
   # this also tests an attribute :uses clause with a single symbol
   virtual_delegate :name, :to => :author, :prefix => true, :type => :string
+  # this tests delegates to named child attribute
+  virtual_delegate :author_name2, :to => "author.name", :type => :string
   # delegate to a polymorphic
   virtual_delegate :description, :to => :current_photo, :prefix => true, :type => :string, :allow_nil => true
 
--- a/spec/db/models.rb
+++ b/spec/db/models.rb
@@ -87,6 +87,10 @@ class Author < VirtualTotalTestBase
     has_attribute?("upper_first_book_author_name") ? self["upper_first_book_author_name"] : first_book_author_name.upcase
   end
 
+  def famous_co_authors
+    book_with_most_bookmarks&.co_authors || []
+  end
+
   # basic attribute with uses that doesn't use a virtual attribute
   def book_with_most_bookmarks
     books.max_by { |book| book.bookmarks.size }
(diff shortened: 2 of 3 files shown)


kbrock:
update:
- fixed spacing cops
- added another test case and better document existing test case. (I'm concerned that future me will make the code more efficient and basically remove the case that we are testing)

spec/db/models.rb
```diff
@@ -95,6 +99,11 @@ def book_with_most_bookmarks
   virtual_attribute :first_book_author_name, :string, :uses => {:books => :author_name}
   # uses another virtual attribute that uses a relation
   virtual_attribute :upper_first_book_author_name, :string, :uses => :first_book_author_name
+  # uses another virtual_attribute that uses a complex relation
+  # Please do not "fix" syntax of the uses here
+  #   actually testing this format with an array: [:bwmb, {:books => co_a}]
+  #   vs "better" format not helping us with tests: {:bwmb => {}, :books => co_a}
```
Fryguy, line 105: The wording here is strange to me - I think this is clearer and more direct

```suggestion
  # NOTE: Please do not change the :uses format here.
  #   This intentionally tests :uses with an array: [:bwmb, {:books => co_a}]
  #   vs a "better" format such as: {:bwmb => {}, :books => co_a}
```

spec/virtual_includes_spec.rb
```diff
@@ -462,6 +487,24 @@
     end
   end
 
+  context ".replace_virtual_field (private)" do
+    # would have prefer target: {:books => :bookmarks} without the extra :books. but here we are
```
Fryguy, line 491: I'm not sure what this comment means.

spec/virtual_includes_spec.rb
```diff
@@ -462,6 +487,24 @@
...
+    it "handles deep includes(:uses => :books => :bookmarks)" do
+      expect(Author.replace_virtual_fields([:book_with_most_bookmarks, :books])).to eq([{:books => :bookmarks}, :books])
+      expect(Author.replace_virtual_fields([{:book_with_most_bookmarks => {}}, {:books => {}}])).to eq([{:books => :bookmarks}, {:books => {}}])
+    end
+
+    # I was a little confused that it uses author (association) instead of authors (table name)
+    # but active record seems ok with this
```
Fryguy, line 498: This is weird in code to me - it's like a self-narrative conversation?  Can we write this in third person or something for other readers?

```suggestion
    # TODO: Need to understand why this uses author (association) instead of authors (table name)
    # and why active record seems ok with this
```

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -69,16 +69,23 @@ def include_to_hash(value)
...
           when Array
-            value.flatten.each_with_object({}) { |k, h| h[k] = {} }
+            value.flatten.reduce({}) do |h, k|
+              if k.kind_of?(Hash)
+                merge_includes(h, k)
+              else
+                h[k] ||= {}
+                h
+              end
+            end
```
Fryguy, line 79: Super-minor, but I'm thikning maybe we shouldn't rely on merge_includes returning the hash?  Then we can DRY up the "return the h hash" part

```suggestion
            value.flatten.reduce({}) do |h, k|
              if k.kind_of?(Hash)
                merge_includes(h, k)
              else
                h[k] ||= {}
              end
              h
            end
```

Fryguy (approved):
Overall LGTM - My only comments are very minor and/or comments about the comments.

I'll hold off merge based on how you want to move forward and when you want to release the gem with this fix.

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -69,16 +69,23 @@ def include_to_hash(value)
...
             {}
           else
             value
           end
         end
 
-        # @param [Hash] hash1
-        # @param [Hash] hash2
+        # @param [Hash] hash1 (typically modified)
+        # @param [Hash] hash2 (not modified)
```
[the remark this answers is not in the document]
kbrock: this is private.

spec/virtual_includes_spec.rb
kbrock: it was trying to say that an extra `:books` comes out.
And that is not optimal but is good enough. Pretty sure rails ignores.

Thanks. Rewrote as a TODO.

spec/virtual_includes_spec.rb
kbrock: yea. Think it was just me grumbling. I deleted

lib/active_record/virtual_attributes/virtual_fields.rb
kbrock: made that change. it squashed it down even further

lib/active_record/virtual_attributes/virtual_fields.rb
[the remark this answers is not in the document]
kbrock: the actual code has changed to always modified / not modified

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -69,23 +69,24 @@ def include_to_hash(value)
           when String, Symbol
             {value => {}}
           when Array
-            value.flatten.each_with_object({}) { |k, h| h[k] = {} }
+            value.flatten.each_with_object({}) do |k, h|
+              merge_includes(h, k)
+            end
```
kbrock, line 74: If we tread the string/symbol case as a special case, it would avoid converting `h` to a hash `h => {}` and avoid the deep_merge() call. But I think what we have is better overall

```suggestion
            value.flatten.each_with_object({}) do |k, h|
              if k.kind_of?(Symbol) || k.kind_of?(String)
                h[k] ||= {}
              else
                merge_includes(h, k)
              end
            end
```

kbrock:
@Fryguy you good with this one?

lib/active_record/virtual_attributes/virtual_fields.rb
kbrock: This case is pretty common.

I added it to the code so it wasn't in the back of my mind.

kbrock:
update:

- ~~symbolize association names so they merge better (most likely not necessary)~~
- added special case/optimization when you are 
- added tests around `:includes => [:virtual_attribute_with_no_includes]`

update:

- pulled all the `to_sym` changes into a separate PR

kbrock:
update:
- brought in the tests that were in the extracted PR (for `to_sym`)

kbrock:
@Fryguy This should be good to go.

I pulled out anything that was not directly related to this PR.

I did put in a ton of edge cases (that are probably a little over kill)

Note: Fixes https://github.com/ManageIQ/manageiq/issues/22526

lib/active_record/virtual_attributes/virtual_fields.rb
kbrock: I implemented this behavior.
If I cared enough about this "regression" to document it here, then I cared enough about it to write up some tests and fix it.

PR #122, opened 2023-06-15
support bind variables in virtual delegate reflections

kbrock:
We were not properly escaping the arel in the association de-reference.

This use case is very particular and will only happen with hard coded strings in associations.

13016ba214dbfd98 rewrote to_sql to wrap in unprepared_statement in one spot but not in this spot. So I guess this continued that effort.

a53d9597d4550345 stated there was a problem but had no reproducer This added specs, but they don't look like they would be able to detect the mentioned issue. So I guess this continued that effort as well.

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_delegates.rb
+++ b/lib/active_record/virtual_attributes/virtual_delegates.rb
@@ -261,7 +261,9 @@ module ActiveRecord
         yield arel if block_given?
 
         # convert arel to sql to populate with bind variables
-        ::Arel::Nodes::Grouping.new(Arel.sql(arel.to_sql))
+        conn = to_ref.klass.connection
+        sql = conn.unprepared_statement { conn.to_sql(arel) }
+        ::Arel::Nodes::Grouping.new(Arel.sql(sql))
       end
 
       # determine table reference to use for a sub query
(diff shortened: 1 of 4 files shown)


kbrock:
update:
- added the `virtual_delegate` under test (it was the whole purpose of this PR)

kbrock:
update:
- removed changes to tests (I wanted `author` and `author2` on the `reload` but think the original way is fine.

lib/active_record/virtual_attributes/virtual_delegates.rb
```diff
@@ -261,7 +261,9 @@ def self.select_from_alias(to_ref, col, to_model_col_name, src_model_id)
         yield arel if block_given?
 
         # convert arel to sql to populate with bind variables
-        ::Arel::Nodes::Grouping.new(Arel.sql(arel.to_sql))
+        conn = to_ref.klass.connection
```
jrafanie, line 264: Is this evaluated at load time via `virtual_delegate :description, :to => :fancy_photo, :prefix => true, :type => :string` or lazy evaluated later?  If it's load time, is this hitting the database at require time?

kbrock: nope. lazy. every time

jrafanie (approved):
LGTM

kbrock:
update:
- fixed test to properly test `fancy_photo`

Fryguy (approved):

PR #123, opened 2023-06-16
Support virtual_totals for habtm

kbrock:
We rewrote virtual_totals to leverage the sql generated by rails: https://github.com/ManageIQ/activerecord-virtual_attributes/pull/52

This was to support `has_many :through`, but it also supports habtm

### DIFF
--- a/spec/db/models.rb
+++ b/spec/db/models.rb
@@ -19,9 +19,11 @@ class Author < VirtualTotalTestBase
   virtual_total :total_books_in_progress, :wip_books
   # same as total_books, but going through a relation with order
   virtual_total :total_ordered_books, :ordered_books
-  # virtual total using through
+  # virtual total using has_many :through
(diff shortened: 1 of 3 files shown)


spec/db/models.rb
```diff
@@ -22,6 +22,7 @@ class Author < VirtualTotalTestBase
   # virtual total using through
   virtual_total :total_bookmarks, :bookmarks
   alias v_total_bookmarks total_bookmarks
+  virtual_total :total_co_books, :co_books
```
Fryguy, line 25: Might be useful to have a comment like the other virtual totals above

```suggestion
  # virtual total using has and belongs to many
  virtual_total :total_co_books, :co_books
```

kbrock:
update:
- fixed comment to explain purpose of the virtual_totals

PR #124, opened 2023-06-16
Arel with binds

kbrock:
Depends:

- [ ] #122 (first commit -- this PR undoes it)

It was necessary for years to convert the virtual attributes over to sql to remove
bind variables.

Since we have successfully added tests around habtm, `has_many :through`, and
relations with string types, it feels our coverage is good enough to trust them
around bind variables.

This removes some string processing / objects created, but more importantly, gets us away from calling `Arel.sql(arel.to_sql)` and just passing `arel`.

WIP: would like to merge a few of these other PRs and run miq cross test suite against this.
currently, I'm leaning towards merging this

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_delegates.rb
+++ b/lib/active_record/virtual_attributes/virtual_delegates.rb
@@ -261,7 +261,9 @@ module ActiveRecord
         yield arel if block_given?
 
         # convert arel to sql to populate with bind variables
-        ::Arel::Nodes::Grouping.new(Arel.sql(arel.to_sql))
+        conn = to_ref.klass.connection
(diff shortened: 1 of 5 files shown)


Fryguy (approved):
Cool - this almost feels like a deletion.  Love it.

Fryguy:
@kbrock Is this tested with multiple Rails versions?  I thought CI did multiple rails versions at least to ensure forward compatbility.

kbrock:
aah. took me a moment.
This only works for 1 version of rails.

When we upgrade rails, we change the version number (to match rails)
and if we want changes across versions, we post the patch to earlier versions

kbrock:
UN-WIP: I was on the edge with this one, but I'm exercising this code in the ui and it is working for me

PR #125, opened 2023-06-16
[WIP] Add support for testing future Rails versions

Fryguy:
@kbrock Please review.

I was thinking that we should run specs for future versions but allow them to fail. This allows us to fix future Rails things on master in a forward compatible way, if possible, hopefully making upgrades simpler.

Additionally, this allows easier switching locally by just passing RAILS_VERSION. Default for RAILS_VERSION if not passed is the expected release version.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,6 +14,9 @@ jobs:
         ruby-version:
         - '2.7'
         - '3.1'
+        rails-version:
+        - '6.1'
+        - '7.0'
     services:
       postgres:
         image: postgres:13
@@ -32,6 +35,7 @@ jobs:
         ports:
         - 3306:3306
     env:
+      RAILS_VERSION: ${{ matrix.rails-version }}
       CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
       # for the pg cli (psql, pg_isready) and rails
       PGHOST: localhost
@@ -50,20 +54,23 @@ jobs:
           bundler-cache: true
         timeout-minutes: 30
       - name: Run SQLite tests
+        continue-on-error: ${{ matrix.rails-version != '6.1' }}
         env:
           DB: sqlite3
         run: bundle exec rake
       - name: Run PostgreSQL tests
+        continue-on-error: ${{ matrix.rails-version != '6.1' }}
         env:
           DB: postgresql
           COLLATE_SYMBOLS: false
         run: bundle exec rake
       - name: Run MySQL tests
+        continue-on-error: ${{ matrix.rails-version != '6.1' }}
         env:
           DB: mysql2
         run: bundle exec rake
       - name: Report code coverage
-        if: ${{ github.ref == 'refs/heads/master' && matrix.ruby-version == '3.1' }}
+        if: ${{ github.ref == 'refs/heads/master' && matrix.ruby-version == '3.1' && matrix.rails-version == '6.1' }}
         continue-on-error: true
         uses: paambaati/codeclimate-action@v3.0.0
         env:
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,6 @@
 
 source "https://rubygems.org"
 
-gem "activerecord", "~> 6.1.4"
 gem "mysql2"
 gem "pg"
 gem "sqlite3"
--- a/README.md
+++ b/README.md
@@ -50,6 +50,11 @@ To test with different database adapters, set the DB environment variable:
     DB=mysql bundle exec rake
     DB=sqlite3 bundle exec rake
 
+To test with different versions of Rails, set the RAILS_VERSION environment variable:
+
+    RAILS_VERSION=6.1 bundle exec rake
+    RAILS_VERSION=7.0 bundle exec rake
+
 ## Contributing
 
 Bug reports and pull requests are welcome on GitHub at https://github.com/ManageIQ/activerecord-virtual_attributes .
--- a/activerecord-virtual_attributes.gemspec
+++ b/activerecord-virtual_attributes.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
 
   spec.require_paths = ["lib"]
 
-  spec.add_runtime_dependency "activerecord", "~> 6.1.0"
+  spec.add_runtime_dependency "activerecord", "~> #{ENV.fetch("RAILS_VERSION", "6.1")}.0"
 
   spec.add_development_dependency "byebug"
   spec.add_development_dependency "database_cleaner-active_record", "~> 2.1"
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -11,6 +11,8 @@ require "active_record/virtual_attributes/rspec"
 require "database_cleaner/active_record"
 require "db-query-matchers"
 
+puts "\e[93mUsing ActiveRecord #{ActiveRecord.version}\e[0m"
+
 Dir['./spec/support/**/*.rb'].sort.each { |f| require f }
 
 RSpec.configure do |config|


Gemfile
```diff
@@ -2,7 +2,6 @@
 
 source "https://rubygems.org"
 
-gem "activerecord", "~> 6.1.4"
```
Fryguy, line 5: This is not necessary because it's defined in the gemspec.

.github/workflows/ci.yaml
```diff
@@ -50,20 +54,23 @@ jobs:
           bundler-cache: true
         timeout-minutes: 30
       - name: Run SQLite tests
+        continue-on-error: ${{ matrix.rails-version != '6.1' }}
```
Fryguy, line 57: Alternative, we can manager the "allowed to fail" combinations by creating a new matrix entry called "allow-failure" and leverage the "include" to set those for specific combos

.github/workflows/ci.yaml
```diff
@@ -14,6 +14,9 @@ jobs:
         ruby-version:
         - '2.7'
         - '3.1'
```
Fryguy, line 16: Side question, but is there a reason we don't have ruby 3.0 here?

spec/support/database.rb
```diff
@@ -2,16 +2,26 @@
 require "active_record"
 
 class Database
+  VALID_ADAPTERS = %w[sqlite3 postgresql mysql2].freeze
+
+  def self.adapter
+    # Handle missing and short-form DB values
+    case ENV['DB']
+    when nil, "sqlite" then ENV['DB'] = "sqlite3"
+    when "pg"          then ENV['DB'] = "postgresql"
```
Fryguy, line 11: I added this for the muscle memory 😉

activerecord-virtual_attributes.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
 
   spec.require_paths = ["lib"]
 
-  spec.add_runtime_dependency "activerecord", "~> 6.1.0"
+  spec.add_runtime_dependency "activerecord", "~> #{ENV.fetch("RAILS_VERSION", "6.1")}.0"
```
kbrock, line 28: I'd prefer the gemspec to be hardcoded.

Even if it is a `"> 6.1"` and a specific version is coded into the `Gemfile`

.github/workflows/ci.yaml
kbrock: I like the matrix having "bail on this version".

That is very clean

.github/workflows/ci.yaml
```diff
@@ -50,20 +54,23 @@ jobs:
...
       - name: Run SQLite tests
+        continue-on-error: ${{ matrix.rails-version != '6.1' }}
         env:
           DB: sqlite3
         run: bundle exec rake
       - name: Run PostgreSQL tests
+        continue-on-error: ${{ matrix.rails-version != '6.1' }}
         env:
-          DB: pg
+          DB: postgresql
```
kbrock, line 64: heh. I've been using driver name for so long I don't notice this anymore.

README.md
```diff
@@ -44,10 +44,16 @@ After checking out the repo, run `bin/setup` to install dependencies. Then, run
 
 To install this gem onto your local machine, run `bundle exec rake install`. To release a new version, update the version number in `version.rb`, and then run `bundle exec rake release`, which will create a git tag for the version, push git commits and tags, and push the `.gem` file to [rubygems.org](https://rubygems.org).
 
+To test with different versions of Rails, set the RAILS_VERSION environment variable:
```
kbrock, line 47: I'm a little nervous having a single `Gemfile` for multiple rails versions

But then you end up with multiple `Gemfile` copies and basically are back at `appraisals`.

For now this looks pretty basic.

.github/workflows/ci.yaml
Fryguy: I was thikning 3.0 because that's what we ship with.

activerecord-virtual_attributes.gemspec
Fryguy: I don't think we can do that, because the gemspec overrides it, I think ... I'll have to see

README.md
Fryguy: Yeah, the gemspec switch is very simple because we only have one gem we are switching on - once you get past that where you need multiple gems, then Appraisals is probably better.  Not worth the overhead though for this one gem.

spec/support/database.rb
```diff
@@ -22,19 +32,45 @@ def setup
...
+  end
+
+  def database_exists?
+    ActiveRecord::Base.establish_connection(@connection_options)
+    !!ActiveRecord::Base.connection
+  end
+
+  def create_database
+    database = @connection_options.delete("database")
+    ActiveRecord::Base.establish_connection(@connection_options)
```
kbrock, line 68: I tend to create the database outside of here. In my mind, the database admin is a different concern than the application itself / the test suite.

Locally I have a `bin/setup` which does a createdb.
It wasn't nailed down so I didn't get it up to snuff.

which creates the DB in much the same way that `.github/workflows/ci.yml` does (or at least used to)
```

psql -c 'create database virtual_attributes;' || echo 'db exists'
mysql -u root -e 'CREATE SCHEMA IF NOT EXISTS 'virtual_attributes';
```

activerecord-virtual_attributes.gemspec
Fryguy: Yeah, you can't do that.  If the gemspec hardcodes it you can't change it later unless we "hack" it the way override_gem does it, but that feels worse.

`"> 6.1"` is bad because then when we try to install it outside of this application, it will force Rails 7 when want it to be Rails 6.

Would this be more acceptable?

```suggestion
  if ENV["RAILS_VERSION"]
    spec.add_runtime_dependency "activerecord", "~> #{ENV["RAILS_VERSION"]}.0"
  else
    spec.add_runtime_dependency "activerecord", "~> 6.1.0"
  end
```

Fryguy: Actually override_gem style might be preferable - I'll investigate that.

spec/support/database.rb
Fryguy: Same, I figured this just covers all bases and gets out of the way.

spec/support/database.rb
Fryguy: Damn I totally didn't see the bin/setup.  It's not being called in CI (I checked ci.yaml for how to set things up).  Ok, if that exists, I might just drop the database setup bit entirely and just doc that in the README.

Fryguy: Ok I removed this DB setup entirely, preferring bin/setup made a cleanup PR for it: #126

.github/workflows/ci.yaml
kbrock: lets do it!

README.md
kbrock: https://github.com/ManageIQ/activerecord-virtual_attributes/pull/106

If we want to support 7.x, lets create another branch.

Up until this week we haven't found many bugs. So just stating we are using 7.0 from now on seemed to make sense. Now with all of these changes I'm not so confident.

Maybe we can get in some big changes now, cut a branch, and then move master over to 7.0?

Or maybe it is time to go back to multiple ruby versions.

Of note, we are currently not supporting SemanticVer here. I'm surprised I can't find the discussion threads we had with NickL, Oleg, you, and maybe LJ? We did not take going away from SemVer lightly.

I've helped support quite a few active record gems. They all tend to have growing pains here and there when the code is too based upon a previous version of active record. I do believe avoiding semantic versioning for this gem makes sense.

activerecord-virtual_attributes.gemspec
kbrock: The gemspec needs to be static.

.github/workflows/ci.yaml
Fryguy: I'll make a separate PR

activerecord-virtual_attributes.gemspec
Fryguy: Yes, I agree - I need to find a way around this, and I want to leave the gemspec alone.

README.md
Fryguy: > If we want to support 7.x, lets create another branch.

Ok, yeah - that seems like the best approach, because if you want to test on that branch, then just switch to that branch.

Fryguy:
After discussion, I'm closing this one.

PR #127, opened 2023-06-19
[WIP] Supports ruby code for default value

kbrock:
Supports ruby code for default value

Problem
=======

```ruby
virtual_delegate :name, :to => other, :default => _("Unknown")
```

The above code runs the localization routine at class load time.
We want a way to run it at runtime.

```ruby
virtual_delegate :name, :to => other, :default => '_("Unknown")'
```

Returns the actual string referenced.

We do not support lambdas:

```ruby
virtual_delegate :name, :to => other, :default => -> { _("Unknown") }
```

We need a way to lazy evaluate the code

Solution
===

This takes inspiration from `Arel.sql("value")`.
You are explicitly telling people that you are using this sql without escaping.

In our case, we are using this ruby code for our default value (and not an escaped string)

```ruby
virtual_delegate :name, :to => other, :default => ActiveRecord::VirtualAttributes.code('_("Unknown")')
```

### DIFF
--- a/lib/active_record/virtual_attributes.rb
+++ b/lib/active_record/virtual_attributes.rb
@@ -4,6 +4,7 @@ require "active_record"
 require "active_record/virtual_attributes/virtual_includes"
 require "active_record/virtual_attributes/virtual_arel"
 require "active_record/virtual_attributes/virtual_delegates"
+require "active_record/virtual_attributes/ruby"
 
 module ActiveRecord
   module VirtualAttributes
@@ -11,6 +12,10 @@ module ActiveRecord
     include ActiveRecord::VirtualAttributes::VirtualIncludes
     include ActiveRecord::VirtualAttributes::VirtualArel
     include ActiveRecord::VirtualAttributes::VirtualDelegates
+
+    def self.ruby(value)
+      Ruby.new(value)
+    end
 
     module Type
       # TODO: do we actually need symbol types?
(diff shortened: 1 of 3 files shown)


Fryguy:
This code kind of scares me for multiple reasons security wise, unless I'm misunderstanding how it works.  2 questions:

1. Why can't we do lambdas?
2. Typically, in our application, database queries return not-localized data (i.e. English/static strings), and the UI's job is to localize it.  In your example we would/should be using `N_('Unknown')`

kbrock:
confused by `N_` [ref](https://github.com/grosser/fast_gettext/blob/master/lib/fast_gettext/translation.rb#L60-L62)

```ruby
virtual_delegate :name, :to => other, :default => N_("Unknown") # returns the above lambda
```

From what I can tell it doesn't seem to do anything. very bizarre.

Fryguy:
`N_("Unknown")` is a delayed translation.  It tells gettext to put "Unknown" into the i18n catalog for translation but doesn't translate it at that time, and then it returns the original English string.  This way, the value will always return English, and it's effectively a constant.

Later code can do `_(record.name)`, and if name returns the default `"Unknown"` English string it will translated at that time.

kbrock:
WIP: we are punting on this behavior
Saying we don't need to support just in time defaults at this time

kbrock:
Hoping to avoid this. may resurrect in the future

PR #132, opened 2023-10-25
Version 6.1.2 notes

kbrock:
Reminder: this is not semver. this needs to be version 6.1.x since this is to support rails 6.1.x

This feels low risk.

---

* Fix bind variables for joins with static strings #124
* Add `virtual_total` for `habtm` #123
* Fix: `:uses` clause now works with an array and nested hashes. #120
* Uses symbols in the `includes()` clause. defined by `virtual_attribute :uses` and virtual_delegate. #128

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ The versioning of this gem follows ActiveRecord versioning, and does not follow 
 
 ## [Unreleased]
 
+## [6.1.2] - 2023-10-26
+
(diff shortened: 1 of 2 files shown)


Fryguy (approved):

Fryguy:
Do we need to add the following to the changelog also?

- #124 
- #122

kbrock:
update:
- added note around bind variables bug fix (both PRs were same issue)

PR #133, opened 2024-01-12
[WIP] Support Rails 7.0

jrafanie:
Allow rails 7 gems in gemspec

see https://github.com/rails/rails/pull/40776 for changes around the preloader

### DIFF
--- a/activerecord-virtual_attributes.gemspec
+++ b/activerecord-virtual_attributes.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
 
   spec.require_paths = ["lib"]
 
-  spec.add_runtime_dependency "activerecord", "~> 6.1.0"
+  spec.add_runtime_dependency "activerecord", ">=6.1.7.6"
 
   spec.add_development_dependency "byebug"
   spec.add_development_dependency "database_cleaner-active_record", "~> 2.1"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,6 +15,9 @@ jobs:
         - '2.7'
         - '3.0'
         - '3.1'
+        rails-version:
+        - '6.1'
+        - '7.0'
     services:
       postgres:
         image: postgres:13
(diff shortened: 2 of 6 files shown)


jrafanie:
@kbrock early call for help whenever you have time to look at the rails 7 failures or guide me on where to start... 🙇

Fryguy:
I had started down this path in https://github.com/ManageIQ/activerecord-virtual_attributes/pull/125, but since this gem follows version matching with Rails, we decided to close that one.  Should this be on a new separate branch?

activerecord-virtual_attributes.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
 
   spec.require_paths = ["lib"]
 
-  spec.add_runtime_dependency "activerecord", "~> 6.1.0"
+  spec.add_runtime_dependency "activerecord", ">=6.1.7.6"
```
Fryguy, line 28: I think this was the main sticking point in #125 and is why I closed it - in production this unlocks all versions of Rails, but this gem is intentionally _not_ semver and matches the version of Rails that it supports directly.  That is, this gem at this version can't (and shouldn't) work with versions >= 6.1.

In my PR I tried to make this part dynamic and only for test, but it turns out you really can't have "dynamic" code in a gemspec because it won't be packaged properly.

kbrock, line 28: agreed, this needs to be:

```suggestion
  spec.add_runtime_dependency "activerecord", "~>7.0.8"
```

Fryguy: and then also drop the 6.1 test suites.

jrafanie: will do this once 6.1/7.0 get feature parity.

kbrock:
Ugh. Joe. I just blew away this branch for you.
Please repush yours

kbrock:
update:
- fixed some cops

(still unsure if I should open another branch or just keep pushing here)

kbrock:
update:
- rubocop spacing fix

jrafanie:
This is being replaced by #150

PR #140, opened 2024-04-26
Fix issue where sqlite3 2.x doesn't want to load with activerecord 6.1

Fryguy:
Additionally, once sqlite3 is downgraded, the latest json 2.7.2 causes issues where OpenStruct is not defined, which worked with json 2.7.1. Only the specs use OpenStruct, so we should explicitly require it.

@kbrock Please review.  This should get master back to green.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,9 +2,9 @@
 
 source "https://rubygems.org"
 
-gem "activerecord", "~> 6.1.4"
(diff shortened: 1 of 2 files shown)


Fryguy:
I think this should also go to the 6-1-stable branch?  I can't remember how we arrange that.

PR #143, opened 2024-06-12
add missing developer dependencies

kbrock:
the build suite depends upon these gems.
Adding them


kbrock:
update:
- cop: sorted gems
- cop: added mfa = true

PR #149, opened 2024-07-24
replace_virtual_field condenses returning includes

kbrock:
`relace_virtual_field` sometimes returns unnecessary tiers of embedded hashes or arrays. E.g. `[[:field]]` instead of `:field`.

6.1 has one fixes, but is by no means exhaustive:

```ruby
association = association.first if association.kind_of?(Array) && association.size == 1
```

This fixes `replace_virtual_field` to not output that value. Since this is at the return of each level, it essentially applied recursively and flattens it significantly.

Added `freeze` in tests to verify that the input parameter is unmodified.

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -33,18 +33,20 @@ module ActiveRecord
         end
 
         def replace_virtual_fields(associations)
-          return associations if associations.blank?
+          return nil if associations.blank?
 
-          case associations
-          when String, Symbol
-            virtual_field?(associations) ? replace_virtual_fields(virtual_includes(associations)) : associations.to_sym
-          when Array
-            associations.collect { |association| replace_virtual_fields(association) }.compact
-          when Hash
-            replace_virtual_field_hash(associations)
-          else
-            associations
-          end
+          ret =
+            case associations
+            when String, Symbol
+              virtual_field?(associations) ? replace_virtual_fields(virtual_includes(associations)) : associations.to_sym
+            when Array
+              associations.filter_map { |association| replace_virtual_fields(association) }
+            when Hash
+              replace_virtual_field_hash(associations)
+            else
+              associations
+            end
+          simplify_includes(ret)
         end
 
         def replace_virtual_field_hash(associations)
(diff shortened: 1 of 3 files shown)


kbrock:
update:
- rubocop: added parens
- renamed `demote_includes` to `simplify_includes`

kbrock:
this is now part of #133 - we can merge here or there

kbrock:
update:
- fixed bug introduced by rebase

jrafanie:
@kbrock I'm not sure what this fixes.   It looks like it fixes a bug but can't envision how this occurs.

> In the rails 7.0 migration, I had [[:field]] that I needed to convert into [:field].

Can you describe the problem you had here?

kbrock:
Hey @jrafanie 

ok, I added a pre-commit to this PR.
First commit shows the output with a double array example.
The second commit shows that this example is fixed.

This over optimizes the list for rails 7, and that is why I added the `Array[klass_associations]` in the preloader - that rubocop did not like.

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -33,18 +33,20 @@ def virtual_field?(name)
         end
 
         def replace_virtual_fields(associations)
-          return associations if associations.blank?
+          return nil if associations.blank?
```
jrafanie, line 36: So, my biggest concern is I thought I remembered this method was used for it's side effect (modifying the associations in place) and for the return value.  I thought I saw it used for one or other other situation so I'm a little concerned that we are now returning nil when it could have been returning `[]` previously.

Are you comfortable with all the call locations handling nils?

kbrock: ## No side effects

There are no side effects on `associations`
We specifically test this method `freezing` the input to ensure this

I'm comfortable with no sided effects.

## return `nil`

There are 3 callers for this method.

I'm comfortable with all 3 call sites.

### `merge_includes`

This has specific tests around `{}`, `[]`, and `nil` input. They all produce the same output.
The `:symbol` vs `[:symbol]` use cases are tested as well, and these produce the same output.

### `construct_join_dependency`

This does not handle a `nil` so there is the `|| {}` to fix.
The tests handle in hashes, arrays, symbols, nil.

### loader

The loader resolves the associations.
There is a `blank?` check to handle a nil.
There is a `Array[]` to handle when it returns a symbol

jrafanie (approved):
Ok, let's get this in and tested with our rails 7 work.

PR #150, opened 2024-07-30
Add rails 7.0 support and drop rails 6.1

kbrock:
Fixed includes and select issues

we're good to go for rails 7.0 with this

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -151,6 +151,30 @@ module ActiveRecord
 
   module Associations
     class Preloader
+      prepend(Module.new {
+        # preloader is called with virtual attributes - need to resolve
+        def call
+          # Possibly overkill since all records probably have the same class and associations
+          # use a cache so we only convert includes once per base class
+          assoc_cache = Hash.new { |h, klass| h[klass] = klass.replace_virtual_fields(associations) }
+
+          # convert the includes with virtual attributes to includes with proper associations
+          records_by_assoc = records.group_by { |rec| assoc_cache[rec.class] }
+          # if these are the same includes, then do the preloader work
+          return super if records_by_assoc.size == 1 && records_by_assoc.keys.first == associations
+
+          # for each of the associations, run a preloader
+          records_by_assoc.each do |klass_associations, klass_records|
+            next if klass_associations.blank?
+
+            Array[klass_associations].each do |klass_association| # rubocop:disable Style/RedundantArrayConstructor
+              # this calls back into itself, but it will take the short circuit
+              Preloader.new(:records => klass_records, :associations => klass_association, :scope => scope).call
+            end
+          end
+        end
+      })
+
       class Branch
         prepend(Module.new {
           # from branched.rb 7.0
@@ -167,6 +191,28 @@ module ActiveRecord
               (h[reflection] ||= []) << record
             end
             h
+          end
+
+          # branched.rb 7.0
+          def preloaders_for_reflection(reflection, reflection_records)
+            reflection_records.group_by do |record|
+              # begin virtual_attributes changes
+              needed_association = record.class.replace_virtual_fields(association)
+              # end virtual_attributes changes
+
+              klass = record.association(needed_association).klass
+
+              if reflection.scope && reflection.scope.arity != 0
+                # For instance dependent scopes, the scope is potentially
+                # different for each record. To allow this we'll group each
+                # object separately into its own preloader
+                reflection_scope = reflection.join_scopes(klass.arel_table, klass.predicate_builder, klass, record).inject(&:merge!)
+              end
+
+              [klass, reflection_scope]
+            end.map do |(rhs_klass, reflection_scope), rs|
+              preloader_for(reflection).new(rhs_klass, rs, reflection, scope, reflection_scope, associate_by_default)
+            end
           end
         })
       end if ActiveRecord.version >= Gem::Version.new(7.0)
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -151,97 +151,19 @@ module ActiveRecord
 
   module Associations
     class Preloader
-      prepend(Module.new {
-        # preloader.rb active record 6.0
-        # changed:
-        # since grouped_records can return a hash/array, we need to handle those 2 new cases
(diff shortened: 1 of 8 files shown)


lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -136,102 +143,70 @@ class Base
...
-            grouped_records(parent, records, polymorphic_parent).flat_map do |reflection, reflection_records|
-              loaders = preloaders_for_reflection(reflection, reflection_records, scope, polymorphic_parent)
-              recs = loaders.flat_map(&:preloaded_records).uniq
-              child_polymorphic_parent = reflection && reflection.respond_to?(:options) && reflection.options[:polymorphic]
-              loaders.concat Array.wrap(child).flat_map { |assoc|
-                preloaders_on assoc, recs, scope, child_polymorphic_parent
-              }
-              loaders
+        # preloader is called with virtual attributes - need to resolve
+        def call
```
jrafanie, line 147: where is `call` being called?

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -136,102 +143,70 @@ class Base
...
               (h[reflection] ||= []) << record
             end
             h
           end
+
+          # branched.rb 7.0
+          def preloaders_for_reflection(reflection, reflection_records)
+            reflection_records.group_by do |record|
+              # begin virtual_attributes changes
+              needed_association = record.class.replace_virtual_fields(association)
```
jrafanie, line 191: is this for fixing a test fix?  This looks like it could "fix" live code with rails 7.  I'm not sure what we weren't testing previously with rails7 and master on virtual attributes to not need this. Is it just a performance fix only, so it doesn't break without it?

jrafanie:
@kbrock in https://github.com/ManageIQ/activerecord-virtual_attributes/compare/master...jrafanie:activerecord-virtual_attributes:allow_rails_7_backup?expand=1 (the backup for the old PR), I was also able to delete `without_virtual_includes`, `apply_join_dependency`, and `calculate` from `virtual_fields.rb`.  Can you see if we can remove them?  See: https://github.com/ManageIQ/activerecord-virtual_attributes/commit/c4a34911f123bbb1cdcc204e9551cfdfc3c397dc

lib/active_record/virtual_attributes/virtual_fields.rb
kbrock: Great question.

Sure, we do explicitly call it in our specs: https://github.com/kbrock/activerecord-virtual_attributes/blob/allow_rails_7/spec/virtual_includes_spec.rb#L542

But most of the time, it is called directly by rails.

From what I can tell, this is mostly called directly from preload_associations via exec_query.

https://github.com/rails/rails/blob/7-0-stable/activerecord/lib/active_record/relation.rb#L825-L832

I did try patching in that caller, but when we ran the preloader directly, this had issues - as you know since you had to skip those 3 tests before (we were since able to unskip).

So this fixes 18 specs (probably 50+ uses) in virtual_includes_specs like:

```
expect(Author.includes(:nick_or_name)).to preload_values(:nick_or_name, author_name)
```

This `includes` value ends up in a Preloader. As you know, it is not valid. So we need to translate from the old includes values to the proper ones.

Now I did try and just update `@associations` value directly, but that only worked if there were no additional levels, so many tests were still failing.

This fix called allows us to get more complicated with our associations.
Coincidently, this recursive-ish solution is similar to the v6 changes where we call into self if the value changed. That was purely coincidental but brought me comfort.

lib/active_record/virtual_attributes/virtual_fields.rb
kbrock: hmm... just tested.
We definitely had this before the preloader changes I did before.
So I think it did more in the past.

Now, it looks like this only fixes the 2 polymorphic tests.

```ruby
it "preloads through polymorphic (polymorphic => virtual_attribute)" do
  books = Book.includes(:author_or_bookmark => :total_books).load
  expect { expect(books.map { |b| b.author_or_bookmark.total_books }).to eq([3, 3, 3]) }.to_not make_database_queries
end
```

```
Association named 'total_books' was not found on Author; perhaps you misspelled it?
```

kbrock:
@jrafanie good catch. We can definitely deleted them. I thought I had.
they must have snuck back in during rebasing.

kbrock:
update:
- rebased
- added comments
- extracted a few PRs that are not directly responsible

update:
- removed extra changes to grouped_records - it had extra changes that were unnecessary. Think the changes to includes fixed this stuff as well.

lib/active_record/virtual_attributes/virtual_fields.rb
jrafanie: ok, that makes sense.  `includes` not working wouldn't break cross repo tests but would break tests here that are specifically looking for query counts and the expected optimizations.  All good here.

jrafanie:
https://github.com/ManageIQ/manageiq/pull/22873 is updated to use this branch for rails 7.0 support, and cross repo is being run against it: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/871.  It's 💚 other than consumption but I'll take a look at that.  It's not a new problem.  It has been failing but because it always fails, I've looked at the other failures first.

```
    1) ManageIQ::Showback::DataRollup#engine update data_rollups should return value of the group with field
       Failure/Error: expect(data_rollup.get_group('CPU', 'average')).to eq([data_rollup.resource.metrics.for_time_range(data_rollup.start_time, data_rollup.end_time).average(:cpu_usage_rate_average).to_s, "percent"])
  
         expected: ["50.5", "percent"]
              got: [50.5, "percent"]
  
         (compared using ==)
       # ./spec/models/data_rollup_spec.rb:226:in `block (4 levels) in <top (required)>'
```

jrafanie:
@kbrock I think we're good to go on moving this forward. I think the test failure in consumption is unrelated to virtual attributes. Looking at this PR one more time.

kbrock:
update:
- rebase master
- updated comments in Association::Preloader - grouping by association rather than class (to avoid `return`ing from within the loop

kbrock:
This has changelog in it.
Do we think we will get in #150, or hold off? (I want to update the `CHANGELOG` to reflect)

jrafanie (approved):
We did a cross repo on this branch and it all passed 💚 

https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/871

PR #153, opened 2024-07-30
Drop attributes_builder

kbrock:
extracted from #150 

in f53272a9 we removed 5.1 support, so no longer needed to override this method Should have deleted this method

I had expected some exclusion for virtual attributes, but stock has been working since 5.2, so letting it slide


jrafanie (approved):
LGTM, this is small enough and make sense... let's get this in now.

PR #156, opened 2024-07-30
Document that in 7.1, this monkey patch is no longer needed

kbrock:
I was so excited to get this fix into 7.1


jrafanie (approved):
Nice!  Yes, let's get this minor change in.

PR #157, opened 2024-07-31
Order specs

kbrock:
2 specs that were missing


Fryguy (approved):

kbrock:
@Fryguy I flubbed this test.

If you remove the virtual attribute from the `select()`, this successfully fails on pg.

I was racking my head why this was working. Yea. Turns out I was right, this doesn't work.

PR #159, opened 2024-08-01
Don't run preloader on double arrays

kbrock:
This is an optimization.

`klass_associations` needs to be an array.
It can be a `nil, symbol, Array` or `Hash`.

The `nil` case is handled by `blank?`.
The `Symbol` and `Hash` are converted to an `Array`
But the `Array` was being converted to a double `Array`.

Which was not horrible, but just calls a loader then another loader.

With this change, the loader only runs through the association once.

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -159,7 +159,7 @@ module ActiveRecord
           records_by_assoc.each do |klass_associations, klass_records|
             next if klass_associations.blank?
 
-            Array[klass_associations].each do |klass_association| # rubocop:disable Style/RedundantArrayConstructor
+            Array.wrap(klass_associations).each do |klass_association|
               # this calls back into itself, but it will take the short circuit
               Preloader.new(:records => klass_records, :associations => klass_association, :scope => scope).call
             end


kbrock:
update:
- comment wording

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -159,7 +159,8 @@ def call
           records_by_assoc.each do |klass_associations, klass_records|
             next if klass_associations.blank?
 
-            Array[klass_associations].each do |klass_association| # rubocop:disable Style/RedundantArrayConstructor
+            klass_associations = Array[klass_associations] unless klass_associations.kind_of?(Array)
+            klass_associations.each do |klass_association| # rubocop:disable Style/RedundantArrayConstructor
```
Fryguy, line 163: @kbrock Why is this using `Array[]` as opposed to `Array()`  (Similar question to the rubocop Array redundant constructor).  I think this does the same thing.

```suggestion
            Array(klass_associations).each do |klass_association|
```

kbrock:
```ruby
Array({:a => :b}) = [[:a, :b]]
Array[{:a => :b}) = [{:a => :b}]
```

Probably want `Array.wrap`, thought that feels like it has gone out of favor. (validate this statement please)

kbrock:
update:
- changed over to `Array.wrap()`
- dropped rubocop comments

lib/active_record/virtual_attributes/virtual_fields.rb
Fryguy: I'm fine with Array.wrap - Ruby is not very consistent, and Array.wrap solves that inconsistency

Fryguy (approved):

PR #161, opened 2024-08-02
Remove sql logger setup from specs

kbrock:
Before
------

I like to put sql logging configuration in supports.
With this code here, I am not able to do that.

After
-----

No more setup code. I can now put my logging code of choice anywhere in `spec/support`

### DIFF
--- a/spec/support/database.rb
+++ b/spec/support/database.rb
@@ -28,12 +28,6 @@ class Database
       I18n.enforce_available_locales = false if I18n.respond_to?(:enforce_available_locales=)
       # I18n.fallbacks = [I18n.default_locale] if I18n.respond_to?(:fallbacks=)
     end
-    log = Logger.new(STDERR)
-    # log = Logger.new('db.log')
-    # log.sev_threshold = Logger::DEBUG
-    log.level = Logger::Severity::UNKNOWN
-    ActiveRecord::Base.logger = log
-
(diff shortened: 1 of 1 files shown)


kbrock:
As mentioned in the seeding pr, I would love for this to be called in a hook in seeds.

spec/support/database.rb
```diff
@@ -30,8 +30,7 @@ def setup
     end
     log = Logger.new(STDERR)
     # log = Logger.new('db.log')
-    # log.sev_threshold = Logger::DEBUG
-    log.level = Logger::Severity::UNKNOWN
+    log.level = ENV["SQL"].to_s.match?(/tr/i) ? Logger::Severity::DEBUG : Logger::Severity::UNKNOWN
```
Fryguy, line 33: I'm just curious about the "tr" check?  Are you expecting "true"?  If so match might be too loose, and perhaps .start_with would be cleaner. But why not just check presence? (I like to pass "1" sometimes)

kbrock: Thanks for this comment.
I deleted this code.

Tested with no supports file, and with a logger in supports. All works as expected for me.

kbrock:
for all you following at home (and future me)

here is the local code I use for debugging:

```ruby
# spec/supports/database.local.logging.rb
log = Logger.new(STDERR)
# log = Logger.new('db.log')
log.level = ENV["SQL"]&.downcase == "true" ? Logger::Severity::DEBUG : Logger::Severity::UNKNOWN
ActiveRecord::Base.logger = log
```

PR #162, opened 2024-08-02
pass all parameters to preloader children

kbrock:
This passes all the parameters from the preloader into the new preloader.

This was causing a problem in manageiq preloader specs.

Unfortunately, I do not fully understand these parameters, I just know we want to pass them all.
And I'm no longer able to reproduce the issue.

But not passing these along seems like a mistake


jrafanie:
> Unfortunately, I do not fully understand these parameters, I just know we want to pass them all.
> And I'm no longer able to reproduce the issue.

hmm, that's odd... it seems like it should break all the time if that was the case

kbrock:
We tend to not pass the `scope` into the preloader. Only the metrics collection will see an issue here.

jrafanie:
LGTM, merging.

PR #163, opened 2024-08-05
Remove test files from packaged gem

kbrock:
Goal
----

I just ran into issues with an exploded gem having a `.rubcop.yml` file.
Seems we shouldn't ship these.

So I removed unnecessary files.

Details
-------

Not sure what is classified as necessary.

- In the lists below, I lumped all `lib/*` files together to make it more readable.
- The `bin/*` files are for development. So I think they are not needed for distribution.
- Do we want to keep the `Rakefile` or `Gemfile`? Left in for now.

Before
------

[".codeclimate.yml",
 ".github/workflows/ci.yaml",
 ".gitignore",
 ".rspec",
 ".rubocop.yml",
 ".rubocop_cc.yml",
 ".rubocop_local.yml",
 ".whitesource",
 ".yamllint",
 "CHANGELOG.md",
 "Gemfile",
 "LICENSE.txt",
 "README.md",
 "Rakefile",
 "activerecord-virtual_attributes.gemspec",
 "bin/console",
 "bin/setup",
 "init.rb",
 "lib/*",
 "renovate.json",
 "seed.rb"]

After
-----

["CHANGELOG.md",
 "Gemfile",
 "LICENSE.txt",
 "README.md",
 "Rakefile",
 "activerecord-virtual_attributes.gemspec",
 "init.rb",
 "lib/*"]

### DIFF
--- a/activerecord-virtual_attributes.gemspec
+++ b/activerecord-virtual_attributes.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
   # Specify which files should be added to the gem when it is released.
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
   spec.files = Dir.chdir(File.expand_path('..', __FILE__)) do
-    `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
+    `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features|bin)/}) || f.match(/^(\.)|renovate.json/) }
   end
 
   spec.require_paths = ["lib"]


jrafanie:
> * The `bin/*` files are for development. So I think they are not needed for distribution.

Just for clarity... some gems ship bin files that are for runtime so the above is only in regards to the current state of virtual attributes.

jrafanie:
@kbrock it might be more helpful to list the files removed from the packaged gem file in addition to the ones that remain

Fryguy:
seed.rb is new to me - feels like that needs to move somewhere else?

kbrock:
yea, seed is a test thing. Good idea - moving

Fryguy:
Merged #164

PR #168, opened 2024-10-07
Rails 7.1 support

jrafanie:
* Support Rails 7.1 only on master
* Add ruby 3.2-3.4 to the test matrix, drop 2.7-3.0.

* Rails 7.1 removed the table_name alias to name so just use name:
See: https://www.github.com/rails/rails/pull/46864

* Rails 7.1 removed the writer for the table_alias so set the instance
variable.  Since they dropped the table_alias writer on the Arel::Table, we're
no longer modifying it so we can drop the dup.
See: https://www.github.com/rails/rails/pull/48927

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_delegates.rb
+++ b/lib/active_record/virtual_attributes/virtual_delegates.rb
@@ -271,11 +271,11 @@ module ActiveRecord
       def self.select_from_alias_table(to_klass, src_relation)
         to_table = to_klass.arel_table
         # if a self join, alias the second table to a different name
-        if to_table.table_name == src_relation.table_name
+        if to_table.name == src_relation.name
           # use a dup to not modify the primary table in the model
           to_table = to_table.dup
           # use a table alias to not conflict with table name in the primary query
-          to_table.table_alias = "#{to_table.table_name}_sub"
+          to_table.instance_variable_set(:@table_alias, "#{to_table.name}_sub")
         end
         to_table
       end
--- a/lib/active_record/virtual_attributes/virtual_delegates.rb
+++ b/lib/active_record/virtual_attributes/virtual_delegates.rb
@@ -272,10 +272,8 @@ module ActiveRecord
         to_table = to_klass.arel_table
         # if a self join, alias the second table to a different name
         if to_table.name == src_relation.name
-          # use a dup to not modify the primary table in the model
-          to_table = to_table.dup
           # use a table alias to not conflict with table name in the primary query
-          to_table.instance_variable_set(:@table_alias, "#{to_table.name}_sub")
+          to_table = to_table.alias("#{to_table.name}_sub")
         end
         to_table
       end
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,9 +12,10 @@ jobs:
     strategy:
       matrix:
         ruby-version:
-        - '2.7'
-        - '3.0'
         - '3.1'
+        - '3.2'
+        - '3.3'
+        - '3.4'
     services:
       postgres:
         image: postgres:13
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@
 
 source "https://rubygems.org"
 
-gem "activerecord", "~>7.0.8"
+gem "activerecord", "~>7.1.5"
 gem "mysql2"
 gem "pg"
 gem "sqlite3", "< 2"


lib/active_record/virtual_attributes/virtual_delegates.rb
```diff
@@ -271,11 +271,11 @@ def self.select_from_alias(to_ref, col, to_model_col_name, src_model_id)
       def self.select_from_alias_table(to_klass, src_relation)
         to_table = to_klass.arel_table
         # if a self join, alias the second table to a different name
-        if to_table.table_name == src_relation.table_name
+        if to_table.name == src_relation.name
           # use a dup to not modify the primary table in the model
           to_table = to_table.dup
           # use a table alias to not conflict with table name in the primary query
-          to_table.table_alias = "#{to_table.table_name}_sub"
+          to_table.instance_variable_set(:@table_alias, "#{to_table.name}_sub")
```
jrafanie, line 278: @kbrock do we need to set a table alias in this way?  I didn't look to see if this was yours or Nick's comment.  Clearly, they don't want to let us set it so I concerned this will change in the future.

kbrock: aah - now I remember

rails uses a manager to manage the names of the tables (so you can do a self join and it all works)

I had been meaning to change this for the next revision

If I remember correctly, this is tied to 3 test failures - the only real failures for rails.

They are real faiulres.
Pretty sure we don't want `name` here.

jrafanie: I'll leave this WIP... feel free to take this change in part or whole and change it while I'm away.

kbrock: I'm putting together a PR for tests that will verify this change

jrafanie: @kbrock do you have suggestions on getting this change in?  I'm not a fan of this change so I'm open to something better.

kbrock: a. Pretty sure this is 100% mine.
b. This code is here for self table joins. So you can use a virtual attribute that is referencing this table.

```ruby
class Model
  has_one :parent, :class_name => Model
  virtual_delegate :name, :to => :parent, :prefix_true, :allow_nil => true
end
```

Without this code, `parent_name` actually returns `self.name`.
Napkin code to express what I mean:

without

```sql
select models.*,
       select (models.name from models where models.id = models.parent_id limit 1)
from models
```

There are 2 `models` table in the query and this is 

vs with

```sql
select models.*,
       select (models_sub.name from models models_sub where models_sub.id = models.parent_id limit 1)
from models
```

kbrock: dumb. I'm sorry:

```ruby
# rails 7.0
alias :table_name :name
```

So this change should work for both 7.0 and 7.1.

lib/active_record/virtual_attributes.rb
```diff
@@ -75,12 +75,22 @@ def virtual_attribute?(name)
         )
       end
 
+      def virtual_attribute_alias?(name)
+        return false unless respond_to?(:attribute_alias?)
+
+        attribute_alias?(name) && virtual_attributes_to_define.keys.include?(name)
```
jrafanie, line 81: this whole method could be simply:

```ruby
def virtual_attribute_alias?(name)
  virtual_attribute_alias_names.include?(name)
end
```

lib/active_record/virtual_attributes.rb
```diff
@@ -75,12 +75,22 @@ def virtual_attribute?(name)
...
+      end
+
       def virtual_attribute_names
-        if respond_to?(:column_names)
-          attribute_names - column_names
-        else
-          attribute_names
-        end
+        names = attribute_names | virtual_attribute_alias_names
+        names -= column_names if respond_to?(:column_names)
```
jrafanie, line 92: @kbrock I believe this is the biggest question for this problem.  What exactly is an attribute and what is not.  It feels like it should be:

```
virtual_attributes_to_define.keys
```

Do you know why we're pulling from attribute_names instead of the list of virtual attributes to define?

kbrock: I think the end desire was to not have a virtual attributes list but rather just have it as a rails attribute with the added `arel` attribute

jrafanie:
@kbrock please take a look...any other changes needed here?

kbrock:
I like this but my issue is we version virtual attributes by rails version. Master will only be for rails 7.1, so I don't want the version matrix logic.

If you could put in only the field name logic I would say to merge

jrafanie:
@kbrock ok, pushed... dropped rails 7.0 for master.  Dropped ruby 2.7-3.0 and added 3.2-3.4.

lib/active_record/virtual_attributes/virtual_delegates.rb
jrafanie: Should be resolved!

kbrock, line 278: can we try `to_table = to_table.alias("#{to_table.table_name}_sub")
?

also, maybe comment out the `to_table` bit to figure out if we even have a test for this

jrafanie: > also, maybe comment out the `to_table` bit to figure out if we even have a test for this

Without setting it, it looks like we have tests that depend on it.  The last one looks to be testing it.


fails with this:

```
  1) ActiveRecord::VirtualAttributes::VirtualDelegates delegates to parent (sql)
     Failure/Error: expect(tcs.map(&:x)).to match_array([nil, 4])

       expected collection contained:  [nil, 4]
       actual collection contained:    [nil, nil]
       the missing elements were:      [4]
       the extra elements were:        [nil]
     # ./spec/virtual_delegates_spec.rb:26:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:40:in `block (3 levels) in <top (required)>'
     # /Users/joerafaniello/.gem/ruby/3.3.6/gems/database_cleaner-core-2.0.1/lib/database_cleaner/strategy.rb:30:in `cleaning'
     # /Users/joerafaniello/.gem/ruby/3.3.6/gems/database_cleaner-core-2.0.1/lib/database_cleaner/cleaners.rb:34:in `block (2 levels) in cleaning'
     # /Users/joerafaniello/.gem/ruby/3.3.6/gems/database_cleaner-core-2.0.1/lib/database_cleaner/cleaners.rb:35:in `cleaning'
     # ./spec/spec_helper.rb:39:in `block (2 levels) in <top (required)>'

  2) ActiveRecord::VirtualAttributes::VirtualDelegates with has_one :parent delegates to child (sql)
     Failure/Error: expect { expect(tcs.map(&:child_col1)).to match_array([nil, tc.id]) }.to_not make_database_queries

       expected collection contained:  [nil, 2]
       actual collection contained:    [nil, nil]
       the missing elements were:      [2]
       the extra elements were:        [nil]
     # ./spec/virtual_delegates_spec.rb:84:in `block (4 levels) in <top (required)>'
     # /Users/joerafaniello/.gem/ruby/3.3.6/gems/activesupport-7.1.5.1/lib/active_support/notifications.rb:259:in `subscribed'
     # /Users/joerafaniello/.gem/ruby/3.3.6/gems/db-query-matchers-0.14.0/lib/db_query_matchers/make_database_queries.rb:75:in `block (2 levels) in <top (required)>'
     # ./spec/virtual_delegates_spec.rb:84:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:40:in `block (3 levels) in <top (required)>'
     # /Users/joerafaniello/.gem/ruby/3.3.6/gems/database_cleaner-core-2.0.1/lib/database_cleaner/strategy.rb:30:in `cleaning'
     # /Users/joerafaniello/.gem/ruby/3.3.6/gems/database_cleaner-core-2.0.1/lib/database_cleaner/cleaners.rb:34:in `block (2 levels) in cleaning'
     # /Users/joerafaniello/.gem/ruby/3.3.6/gems/database_cleaner-core-2.0.1/lib/database_cleaner/cleaners.rb:35:in `cleaning'
     # ./spec/spec_helper.rb:39:in `block (2 levels) in <top (required)>'

  3) ActiveRecord::VirtualAttributes::VirtualDelegates with has_one :parent uses table alias for subquery
     Failure/Error: expect(sql).to match(/["`]test_classes_[^"`]*["`][.]["`]col1["`]/i)

       expected "SELECT \"test_classes\".\"id\", \"test_classes\".\"col1\", ((SELECT \"test_classes\".\"col1\" FROM \...test_classes\".\"col1\" = \"test_classes\".\"id\" LIMIT 1)) AS \"child_col1\" FROM \"test_classes\"" to match /["`]test_classes_[^"`]*["`][.]["`]col1["`]/i
       Diff:
       @@ -1 +1 @@
       -/["`]test_classes_[^"`]*["`][.]["`]col1["`]/i
       +"SELECT \"test_classes\".\"id\", \"test_classes\".\"col1\", ((SELECT \"test_classes\".\"col1\" FROM \"test_classes\" WHERE \"test_classes\".\"col1\" = \"test_classes\".\"id\" LIMIT 1)) AS \"child_col1\" FROM \"test_classes\""

     # ./spec/virtual_delegates_spec.rb:92:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:40:in `block (3 levels) in <top (required)>'
     # /Users/joerafaniello/.gem/ruby/3.3.6/gems/database_cleaner-core-2.0.1/lib/database_cleaner/strategy.rb:30:in `cleaning'
     # /Users/joerafaniello/.gem/ruby/3.3.6/gems/database_cleaner-core-2.0.1/lib/database_cleaner/cleaners.rb:34:in `block (2 levels) in cleaning'
     # /Users/joerafaniello/.gem/ruby/3.3.6/gems/database_cleaner-core-2.0.1/lib/database_cleaner/cleaners.rb:35:in `cleaning'
     # ./spec/spec_helper.rb:39:in `block (2 levels) in <top (required)>'

Finished in 0.74258 seconds (files took 0.43978 seconds to load)
328 examples, 3 failures, 1 pending

Failed examples:

rspec ./spec/virtual_delegates_spec.rb:22 # ActiveRecord::VirtualAttributes::VirtualDelegates delegates to parent (sql)
rspec ./spec/virtual_delegates_spec.rb:80 # ActiveRecord::VirtualAttributes::VirtualDelegates with has_one :parent delegates to child (sql)
rspec ./spec/virtual_delegates_spec.rb:89 # ActiveRecord::VirtualAttributes::VirtualDelegates with has_one :parent uses table alias for subquery
```

jrafanie: ok, pushed changes to use the public table alias API.

jrafanie:
[cross repo](https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/937) with core https://github.com/ManageIQ/manageiq/pull/23225/files is 💚

jrafanie:
I think this is ready to go... https://github.com/ManageIQ/activerecord-virtual_attributes/tree/7-0-stable has been cut.

PR #171, opened 2024-11-04
Drop openstruct

kbrock:
openstruct is no longer in the standard library
Could have gone with using `double()` but then it is just swapping one mock for another.
Since we only use it in a few spots, just removed it.


jrafanie (approved):
🎉

PR #174, opened 2025-01-21
Pin concurrent-ruby to < 1.3.5 until we're on rails 7.1

jrafanie:
See core PR:
https://github.com/ManageIQ/manageiq/pull/23310


Fryguy:
Is master Rails 7.0, or do we need a 7.0-stable branch here?

PR #177, opened 2025-02-19
Add tests for delegations to same model

kbrock:
I don't think we are testing joining a model to itself.

It feels contrived, so if you can think of a relationship that makes more sense, please share.

### DIFF
--- a/spec/db/models.rb
+++ b/spec/db/models.rb
@@ -5,6 +5,9 @@ class VirtualTotalTestBase < ActiveRecord::Base
 end
 
 class Author < VirtualTotalTestBase
+  # basically a :parent_id relationship
+  belongs_to :teacher, :foreign_key => :teacher_id, :class_name => "Author"
+  has_many :students, :foreign_key => :teacher_id, :class_name => "Author"
   has_many :books
   has_many :ordered_books,   -> { ordered },   :class_name => "Book"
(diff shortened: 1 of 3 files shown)


Fryguy:
I can't think of any other relationship - Teacher/Student works for me

spec/db/models.rb
```diff
@@ -36,6 +39,9 @@ class Author < VirtualTotalTestBase
   virtual_sum :sum_recently_published_books_rating, :recently_published_books, :rating
   virtual_delegate :description, :to => :current_photo, :prefix => true, :type => :string
   virtual_delegate :description, :to => :fancy_photo, :prefix => true, :type => :string
+  # delegate to parent relationship
+  virtual_delegate :name, :to => :teacher, :prefix => true, :type => :string
```
Fryguy, line 43: although this is weird, because now you have author#name but it's actually teacher_name?

kbrock: the fact that you responded and had an EDIT EDIT just further emphasizes that I did not do a good job with the relationship/naming

kbrock:
update:
- rebased

I did verify that the alias code is tested elsewhere, but this one is so explicit, I think I would like to keep

Fryguy (approved):

PR #182, opened 2025-05-28
Drop rails version 7.0 patch

kbrock:
We're running rails 7.1, drop patch for older versions of rails.

(and we're in the process of upgrading to rails 7.2)

UPDATE:
oops - I jumped the gun on this one. We're going 7.0 -> 7.1
UPDATE 2:
heh - no, this is right. these fields had not been changed when we upgraded to 7.1


Fryguy:
Just for my clarity, since we have the 7-0-stable, this effectively makes makes master 7-1 only?

jrafanie:
> Just for my clarity, since we have the 7-0-stable, this effectively makes makes master 7-1 only?

yeah, I think that's the idea.

kbrock:
@Fryguy Also of note. version 7.1 of our product only targets 7.1.

Once we target 7.2, master will only target 7.2

Fryguy (approved):
LGTM - @jrafanie Please also review.

PR #186, opened 2025-06-13
Early includes and arel definition

kbrock:
@jrafanie this was the next step

Depends:
- [x] #185 (first commit)

### DIFF
--- a/lib/active_record/virtual_attributes.rb
+++ b/lib/active_record/virtual_attributes.rb
@@ -103,8 +103,8 @@ module ActiveRecord
 
       def define_virtual_attribute(name, cast_type, uses: nil, arel: nil)
         attribute_types[name.to_s] = cast_type
-        define_virtual_include(name, uses) if uses
-        define_virtual_arel(name, arel) if arel
+        define_virtual_include(name, uses)
+        define_virtual_arel(name, arel)
       end
     end
   end
--- a/lib/active_record/virtual_attributes.rb
+++ b/lib/active_record/virtual_attributes.rb
@@ -98,10 +98,6 @@ module ActiveRecord
 
           define_virtual_attribute(name, type)
         end
-
-        virtual_delegates_to_define.each do |method_name, (method, options)|
-          define_virtual_delegate(method_name, method, options)
-        end
       end
 
       def define_virtual_attribute(name, cast_type)
--- a/lib/active_record/virtual_attributes.rb
+++ b/lib/active_record/virtual_attributes.rb
@@ -56,12 +56,15 @@ module ActiveRecord
         virtual_attribute(name, type, **options)
       end
 
-      def virtual_attribute(name, type, **options)
+      def virtual_attribute(name, type, uses: nil, arel: nil, **options)
         name = name.to_s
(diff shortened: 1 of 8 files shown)


kbrock:
@jrafanie I think I'm ok with the code climate error

Lets merge it

lib/active_record/virtual_attributes.rb
```diff
@@ -91,18 +94,12 @@ def load_schema!
           type = type.call if type.respond_to?(:call)
           type = ActiveRecord::Type.lookup(type, **options.except(:uses, :arel)) if type.kind_of?(Symbol)
 
-          define_virtual_attribute(name, type, **options.slice(:uses, :arel))
-        end
-
-        virtual_delegates_to_define.each do |method_name, (method, options)|
-          define_virtual_delegate(method_name, method, options)
+          define_virtual_attribute(name, type)
```
jrafanie, line 97: Nice, this is much clearer with the uses and arel handled elsewhere.  This method has a single responsibility now.

lib/active_record/virtual_attributes/virtual_delegates.rb
```diff
@@ -9,10 +9,6 @@ module VirtualAttributes
     module VirtualDelegates
       extend ActiveSupport::Concern
 
-      included do
-        class_attribute :virtual_delegates_to_define, :instance_accessor => false, :default => {}
-      end
-
```
jrafanie, line 15: 🔥 🗑️ 🚽  => 🎉

lib/active_record/virtual_attributes/virtual_delegates.rb
```diff
@@ -46,38 +46,21 @@ def virtual_delegate(*methods)
               options[:to] = to
             end
 
+            unless (to_ref = reflection_with_virtual(to.to_s))
+              raise ArgumentError, "Delegation needs an association. Association #{to} does not exist"
```
jrafanie, line 50: Similar question here as as the type... should we deprecate on the current release and do this raise in the new version?

kbrock:
@miq-bot cross_repo_test /all

lib/active_record/virtual_attributes.rb
kbrock: I was debating dropping the `options` here, since it is only used very deep in the `Type.lookup` discovery code, and as far as I know, we don't use this. But maybe array or something leverages this, so punting and leaving it in

lib/active_record/virtual_attributes/virtual_delegates.rb
Fryguy: +1 for dropping it and making it explicit.  If in the future something comes along where we want it dynamic we can add it back in later, hopefully better.

kbrock: Sorry I missed your question Joe.
The `to_ref` has always been required. This just moved the check from one method to another.

kbrock:
update:
- changed individual assignments to define_virtual_attribute
update:
- fixed rubocop hashrocket syntax

PR #187, opened 2025-06-13
V72

kbrock:
https://github.com/rails/rails/commit/e0a55b038f7f2f50d1467876558be183be6cedaa

While we do not leverage the `attribute` api, we are strongly influence by it.
We do not want a default value record, which modifies select(*) calls and others.
We do to add the values to attribute_type.

Before
------

The values are stored in `attributes_to_define_after_schema_loads` at class load time,
and in load_schema, are then used to generate entries for `DefaultValue` and `attribute_type`.

After
-----

The values are stored as `ModifyAttribute` entries.
Those values then are used to populate default values and attribute_type on demand.

Our Change
----------

Previous PRs removed arel and uses logic so load_schema only manipulated attribute_type.
This changes like the rails change to populate (modify) attribute_type on demand.

depends:
- [x] #188
- [x] #186
- [x] #185

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,15 @@ The versioning of this gem follows ActiveRecord versioning, and does not follow 
 
 ## [Unreleased]
 
+## [7.2.0.0] - 2025-06-25
+
+* virtual_delegate requires type [#185](https://github.com/ManageIQ/activerecord-virtual_attributes/pull/185)
+* Rails 7.2 support [#187](https://github.com/ManageIQ/activerecord-virtual_attributes/pull/187)
+
 ## [7.1.2] - 2025-06-20
 
 * Introduce virtual_has_many :through [#191](https://github.com/ManageIQ/activerecord-virtual_attributes/pull/191)
+* Use kwargs for virtual_delegate and virtual_column [#190](https://github.com/ManageIQ/activerecord-virtual_attributes/pull/190)
 
 ## [7.1.1] - 2025-06-18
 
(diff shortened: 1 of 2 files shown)


lib/active_record/virtual_attributes.rb
```diff
@@ -85,16 +85,13 @@ def virtual_attribute_names
...
-          type = ActiveRecord::Type.lookup(type, **options.except(:uses, :arel)) if type.kind_of?(Symbol)
-
-          define_virtual_attribute(name, type)
+      def attribute_types
+        @attribute_types || super.tap do | hash|
+          virtual_attributes_to_define.each do |name, (type, options)|
+            type = type.call if type.respond_to?(:call)
+            type = ActiveRecord::Type.lookup(type, **options.except(:uses, :arel)) if type.kind_of?(Symbol)
+            hash[name] = type
+          end
```
jrafanie, line 94: wow, this is the only change for 7.2 after those others?  Crazy!

kbrock: So #186 is all the changes. So I wanted to run cross repo on that and make those changes

kbrock:
update:
- rebased
- changed version to 7.2.0.0 (SemVer)

jrafanie:
> I put in version changes in here, and put in rubocop changes (8 commits)
> I can start pulling out PRs if you need. I'd prefer to just get this out.
> And start working on rails 8.0 :)

yeah, I think there's some value with doing the attribute_types change, explain why it's needed for rails 7.2, and then have the version bump in a second commit or split them entirely.  Thoughts @kbrock ?

jrafanie (approved):
LGTM

PR #189, opened 2025-06-17
fix rubocops

kbrock:
Still have a bunch of rubocops in the spec due to the way we create the `TestClass`.
But this should resolve most other issues.

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_arel.rb
+++ b/lib/active_record/virtual_attributes/virtual_arel.rb
@@ -117,4 +117,3 @@ module ActiveRecord
     end
   end
 end
-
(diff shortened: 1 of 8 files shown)


lib/active_record/virtual_attributes/virtual_total.rb
```diff
@@ -116,4 +116,4 @@ def virtual_aggregate_arel(reflection, method_name, column)
   end
 end
 
-ActiveRecord::Base.include VirtualAttributes::VirtualTotal
+ActiveSupport.on_load(:active_record) { include VirtualAttributes::VirtualTotal }
```
Fryguy, line 119: Is the block in the scope of ActiveRecord::Base?  (How does this work?)

kbrock: ooh. good point. That is what Rubocop asked me to change, but I see what you mean.

Not totally sure how this works, I'm going to roll this back.

kbrock:
update:
- rolled back `ActiveSupport.on_load(:active_record)`

- update:
-  added `Rails/ApplicationRecord` ignores
- dropped `Naming/PredicateMethod` ignores

PR #191, opened 2025-06-19
Introduce delegation to virtual_has_many

kbrock:
before
------

We want to define a `virtual_has_many` which is based upon a `has_many` on a parent class.
We want this association to be discoverable and visible in the api.

```ruby
  virtual_delegate :a, :to => "b.c"
  virtual_delegate :j, :to => :k, :prefix => true

  virtual_delegate :cloud_volume_types,
                   :to => "ext_management_system.cloud_volume_types",
                   :allow_nil => true, :default => []
```

The problem is `virtual_delegate` was written to reference a sql friendly attribute on the other side of a sql friendly relation. The delegate above is a relation on a ruby relation. People were using `virtual_delegate` because it provided a `default`.

after
-----

```ruby
  virtual_has_many :a, :through => :b, :source => :c
  virtual_has_many :k_j, :through => :k, :source => :j

  virtual_has_many :cloud_volume_types, :through => :ext_management_system
```

This produces the following ruby:
```ruby
  def a   ; b&.c || [] ; end
  def k_j ; k&.j || [] ; end
  def cloud_volume_types ; ext_management_system&.cloud_volume_types || [] ; end
```

dev notes
---------

The delegate method defined does have some traces of virtual_attributes, namely the `has_attribute?` call, but this seems preferred over re-implementing this method

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ The versioning of this gem follows ActiveRecord versioning, and does not follow 
 
 ## [Unreleased]
 
+## [7.1.2] - 2025-06-20
+
+* Introduce virtual_has_many :through [#191](https://github.com/ManageIQ/activerecord-virtual_attributes/pull/191)
+
 ## [7.1.1] - 2025-06-18
 
 * Deprecate virtual_delegate without a type [#188](https://github.com/ManageIQ/activerecord-virtual_attributes/pull/188)
@@ -115,7 +119,8 @@ The versioning of this gem follows ActiveRecord versioning, and does not follow 
 * Initial Release
 * Extracted from ManageIQ/manageiq
 
-[Unreleased]: https://github.com/ManageIQ/activerecord-virtual_attributes/compare/v7.1.1...HEAD
+[Unreleased]: https://github.com/ManageIQ/activerecord-virtual_attributes/compare/v7.1.2...HEAD
+[7.1.2]: https://github.com/ManageIQ/activerecord-virtual_attributes/compare/v7.1.1...v7.1.2
 [7.1.1]: https://github.com/ManageIQ/activerecord-virtual_attributes/compare/v7.1.0...v7.1.1
 [7.1.0]: https://github.com/ManageIQ/activerecord-virtual_attributes/compare/v7.0.0...v7.1.0
 [7.0.0]: https://github.com/ManageIQ/activerecord-virtual_attributes/compare/v6.1.2...v7.0.0
--- a/lib/active_record/virtual_attributes/version.rb
+++ b/lib/active_record/virtual_attributes/version.rb
@@ -1,5 +1,5 @@
 module ActiveRecord
   module VirtualAttributes
-    VERSION = "7.1.1".freeze
+    VERSION = "7.1.2".freeze
   end
 end


spec/db/models.rb
```diff
@@ -135,6 +135,8 @@ class Book < VirtualTotalTestBase
   has_many :photos, :as => :imageable, :class_name => "Photo"
   has_one :current_photo, -> { all.merge(Photo.order(:id => :desc)) }, :as => :imageable, :class_name => "Photo"
 
+  virtual_has_many :books, :to => :author, :prefix => true
```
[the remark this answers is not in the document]
kbrock: ok. these are tested in the delegate side.
Adding tests in the relation side makes some sense.
I started to run out of options/names here.

`:to` with a prefix doesn't make much sense. I mean it will work but it is strange.
The purpose of the `:to` with a period is to make the caller's method name different from the target's method name. The prefix is similar, tacking on the prefix to the caller's name. But if you are defining the `:to` with a period, then you may as well just define the method name correctly in the first place. And a `:to` with a period and a prefix of true is just plain strange.

kbrock:
update:
- change `:to` method to `:through` and `:source` to follow active record conventions
- converted relations to use kwargs

update:
- fixed rubocop space warning

Fryguy:
That came out so nice. 🎉

Fryguy:
@kbrock Did you forget to push the changes for hany_many through? I see the other one.

kbrock:
> Oh wait, I misread, I assumed the change would cause determine_method_names to either go away or be simplified. Do we still need the . processing in that method anymore?

Yea, so `define_delegate` still uses `:to => :relation` or `:to => "relation.column"`

spec/db/models.rb
```diff
@@ -135,6 +135,8 @@ class Book < VirtualTotalTestBase
   has_many :photos, :as => :imageable, :class_name => "Photo"
   has_one :current_photo, -> { all.merge(Photo.order(:id => :desc)) }, :as => :imageable, :class_name => "Photo"
 
+  virtual_has_many :author_books, :through => :author, :source => :books
```
Fryguy, line 138: Can we get another test for :through without source?  I assume that would be something like

```
  virtual_has_many :books, :through => :author
```

lib/active_record/virtual_attributes/virtual_reflections.rb
```diff
@@ -9,26 +9,24 @@ module ClassMethods
...
+        def virtual_has_many(name, uses: nil, source: nil, through: nil, **options)
           define_method(:"#{name.to_s.singularize}_ids") do
             records = send(name)
             records.respond_to?(:ids) ? records.ids : records.collect(&:id)
           end
-          uses = options.delete(:uses)
           reflection = ActiveRecord::Associations::Builder::HasMany.build(self, name, nil, options)
-          add_virtual_reflection(reflection, name, uses, options)
+          add_virtual_reflection(reflection, name, uses)
+          define_delegate(name, source || name, :to => through, :allow_nil => true, :default => []) if through
```
Fryguy, line 24: Should this method return the virtual reflection like it did before because now it's changed to return the delegate (I presume add_virtual_reflection returns the reflection).

Can we define the delegate before the add_virtual_reflection? If so, we can just swap these 2 lines.

spec/db/models.rb
kbrock: Yea, having trouble thinking up this one. Especially without a source

A book has many books?

kbrock: I really didn't like it but...
Was better than anything I could come up with. - I added it with tests

kbrock:
update:
- kwargs for virtual_has_many and friends
- 

update:
- error on prefix and `to` with a period.
- virtual_has_many returns a delegate
- testing virtual_has_many :through (without :source)
- tests for determine_method_names (yes private, but had lots of questions)

update:
- changed test from {} to use a block

update:
- drop unneeded `to_s`

PR #192, opened 2025-06-24
[WIP] Attribute delegation

kbrock:
Here is an alternative to `virtual_delegates`.

It is essentially the same thing, but using `virtual_attribute` instead of `virtual_delegate`

I like `virtual_delegate` and find it makes sense to me. But here you go.

Diff will show virtual_attribute delta.

```ruby
  virtual_delegate :description, :to => :current_photo, :prefix => true, :type => :string
  virtual_attribute :current_photo_description, :string, :through => :current_photo, :source => :description

  virtual_delegate :blurb, :to => :author, :type => :string
  virtual_attribute :blurb, :string, :through => :author
```


lib/active_record/virtual_attributes.rb
```diff
@@ -56,13 +56,23 @@ def virtual_column(name, type:, **options)
...
           virtual_attributes_to_define.merge(name => [type, options])
 
+        if through
+          define_delegate(name, source, :to => through, :allow_nil => true, :default => default)
+
+          unless (to_ref = reflection_with_virtual(through))
+            raise ArgumentError, "Delegation needs an association. Association #{to} does not exist"
+          end
+
+          arel = virtual_delegate_arel(source, to_ref)
```
Fryguy, line 73: Since this defines `arel`, if `arel` and `through` are both passed, should that be an ArgumentError?

kbrock: I'm thinking passing `arel` should be allowed, though I agree it defeats the purpose of the `:through`

```diff
- arel = virtual_delegate_arel(source, to_ref)
+ arel ||= virtual_delegate_arel(source, to_ref)
```

kbrock:
@Fryguy At a high level, does `virtual_attribute :x, :through` work for you?

Just as `virtual_has_many :through` ended up having a simple solution, this one turned out easier than I thought it would.

As a long term plan, do we think we want to drop `virtual_delegate`?
The code is greatly based upon `delegate`.

Fryguy:
> @Fryguy At a high level, does virtual_attribute :x, :through work for you?

Yes, I think I like `virtual_attribute :through` better than `virtual_delegate` overall from a clarity standpoint.  I read `virtual_attribute` as defining an attribute, and then without `:through` it's a normal attribute and with `:through`, it's through a relationship.  This combines the attribute+delegation concepts into a single thing in my mind which is easier to read, whereas `virtual_delegate` I have to remember that it's also creating an attribute.

kbrock:
update:
- rebased to include 7.2.0.1 changes
- deprecated virtual_delegate
- comments that `source` can not be verified

I went ahead and deprecated `virtual_delegate` in this PR.
While there are 50 instances of virtual_delegate in manageiq, it is only in 2 repos, so should be quick to make the change.

lib/active_record/virtual_attributes/virtual_delegates.rb
```diff
@@ -15,6 +15,8 @@ module ClassMethods
         #
 
         def virtual_delegate(*methods, to:, type:, prefix: nil, allow_nil: nil, default: nil, uses: nil, **options) # rubocop:disable Naming/MethodParameterName
+          ActiveRecord::VirtualAttributes.deprecator.warn("virtual_delegate is now deprecated. use virtual_attribute :through instead", caller_locations)
+
```
Fryguy, line 19: GitHub won't let me comment below but on lines 33-38 below should this method be changed to go through `virtual_attribute :through`? I thikn you can drop the `define_delegate` line and the bit to `raise ArgumentError, "Delegation...` because those are already in virtual_attribute :through.

Fryguy: (Can be in a follow up PR - I just noticed the duplication)

kbrock: I didn't want to work too hard for this one.
The goal is to drop this method all together.

I'm not sure exactly how to invert, I'll look into this.
But do we want to spend the $$$ to do this?

lib/active_record/virtual_attributes.rb
```diff
@@ -56,13 +56,26 @@ def virtual_column(name, type:, **options)
         virtual_attribute(name, type, **options)
       end
 
-      def virtual_attribute(name, type, uses: nil, arel: nil, **options)
+      def virtual_attribute(name, type, through: nil, uses: through, arel: nil, source: name, default: nil, **options)
```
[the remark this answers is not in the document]
kbrock: If someone passes `uses`, and not `through`, then that is an existing `virtual_attributes` use case. For that case, it is expected that a ruby method will be defined that references the association mentioned by the `uses`. e.g.: [ems.rb:25](https://github.com/ManageIq/manageiq/blob/master/app/models/ext_management_system.rb#L253)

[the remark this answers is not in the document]
Fryguy: right, that's not what i'm saying. if someone pass `:uses => :bar, :through => :foo`, then `:uses` won't have `:foo` in it, which I think is a problem.  Alternatively, I think you could tack in the through later on with something roughly like

```ruby
if through
  if uses
    uses = Array(uses) << through
  else
    uses = through
  end
...

lib/active_record/virtual_attributes/virtual_delegates.rb
Fryguy: Doesn't really matter to me - I was trying to reduce the duplication, but yeah if it's going away quickly, then it really doesn't matter.

kbrock: We tend to not specify `uses`.

- With `arel`: 4 of the 45 specify `uses`.
- Without `arel`: 157 of the 371 specify `uses`.
- With delegate: 2 of the 48 specify `uses`.

Without `arel` manually traverse associations, so I expected it to be a higher percentage. Delegates traverse another object, so I wondered if that would be zero. 

The use case we are looking at is the delegate. Since 5% use the `to` and not the `uses`, I'd prefer to keep the `(uses || to)` in there.

The 2 delegate with uses that are similar: `:to => :hardware, :uses => {:hardware => :disks}`. Fun fact: The `:uses` was not necessary a few years ago. We introduced it when we stopped introspecting the target class. (The "Constant" problem race condition that Joe fixed)

PR #193, opened 2025-06-25
Nitpicks

kbrock:
1. rubocop around preloaders_for_reflection. this is a rails method and we don't want to modify them where possible
2. virtual_has_many defaults the `source` rather than modifying it in the code
3. rubocop `__FILE__` => `__dir__`
4. add delegate with no `prefix` declared.


jrafanie (approved):
LGTM

PR #194, opened 2025-07-10
[wip] use rails 8.0

kbrock:
Started as a litmus test. But now I'm wondering

I'd be up for supporting 7.2 and 8.0 branches
May want to get the `virtual_attribute :through` changes into 7.2.1 before merging this / creating another branch

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,8 +12,6 @@ jobs:
     strategy:
       matrix:
         ruby-version:
-        - '3.1'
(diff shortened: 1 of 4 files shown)


jrafanie:
Yeah, I was just considering having 7.2 and 8.0 in the matrix with 8.0 as allowed failures if needed.

jrafanie:
yeah, on the main branch, we should certainly try to support the current supported rails and any newer ones we're trying to get to green with allow failures or something similar.

PR #195, opened 2025-07-18
Fix nil :uses issue when included through a polymorphic

kbrock:
When you `includes(:polymorphic => :attr)` and the `:attr` does not have a `:uses`,
this threw an exception.

Reason: Rails 7.2 added a `to_sym` in `reflect_on_association()`,
since `uses[:attr] = nil`, calling `reflect_on_association(nil)` had issues:

```
     Failure/Error: reflection = record.class._reflect_on_association(association)

     NoMethodError:
       undefined method `to_sym' for nil
     # /Users/kbrock/.gem/ruby/3.3.5/gems/activerecord-7.2.2.1/lib/active_record/reflection.rb:127:in `_reflect_on_association'
     # ./lib/active_record/virtual_attributes/virtual_fields.rb:189:in `block in grouped_records'
```

refs:
- https://github.com/rails/rails/issues/5172
- https://github.com/ManageIQ/manageiq-api/pull/1284
- https://github.com/ManageIQ/manageiq/issues/23518

### DIFF
--- a/lib/active_record/virtual_attributes/virtual_fields.rb
+++ b/lib/active_record/virtual_attributes/virtual_fields.rb
@@ -183,6 +183,7 @@ module ActiveRecord
             source_records.each do |record|
               # begin virtual_attributes changes
               association = record.class.replace_virtual_fields(self.association)
+              next if association.nil?
               # end virtual_attributes changes
 
               reflection = record.class._reflect_on_association(association)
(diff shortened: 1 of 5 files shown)


kbrock:
update:
- rubocop deleted extra empty line at block body end

lib/active_record/virtual_attributes/virtual_fields.rb
```diff
@@ -183,6 +183,7 @@ def grouped_records
             source_records.each do |record|
               # begin virtual_attributes changes
               association = record.class.replace_virtual_fields(self.association)
+              next if association.nil?
```
jrafanie, line 186: haha, all that for this.  Nice and simple.

PR #197, opened 2025-07-23
drop unnecessary load_schema statements

kbrock:
The load_schema are not doing anything.
Then again, they are not hurting anyone.

I just added this b/c we were debugging and researching the connection pool while using the 

- virtual_attribute?:
  - has attribute loads the schema if necessary
  - column_for_attribute loads the schema if necessary
- attribute_supported_by_sql?
  - has_attribute loads the schema if necessary
  - virtual_attribute? loads the schema indirectly (by 2 internal methods)
  - _virtual_arel is populated on class load, not schema load
- virtual_includes
  - _virtual_includes is populated on class load, not schema load


kbrock:
update:
- updated commit comment

jrafanie:
Should we do a release with this change so we can verify it works in cross repo?

Fryguy:
Or we can cross-repo with master

PR #204, opened 2025-09-12
[WIP] [POC] Implement Condensed VA syntax

### LINKED ISSUE #203: Support condensed virtual_attribute definitions
I'm trying to guage what people think of this new syntax.

1. It has the ruby and sql definitions together rather than scattered.
2. It has 2 blocks (arel and ruby) back to back.
3. It auto generates the "use attribute from `select()`" line. i.e.:
   ```return
   return self["archived"] if  has_attribute?("archived")
   ```
4. The "use attribute if from `select()`" is missing all over our code, so that is a win.
5. The ruby method will be defined using metadata rather than ruby. (something to be said for plain old ruby)
6. This can be added incrementally.

NOTE: If you look in the code, the `t.grouping()` adds parens around the SQL. This is no longer necessary and going through this will allow us to remove it.

Before
======

```ruby
  def archived?
    return self["archived"] if  has_attribute?("archived")

    ems_id.nil? && storage_id.nil?
  end
  alias_method :archived, :archived?

  virtual_attribute :archived, :boolean, :arel => (lambda do |t|
    t.grouping(t[:ems_id].eq(nil).and(t[:storage_id].eq(nil)))
  end)
```

After
=====

```ruby
  virtual_attribute :archived, :boolean, :arel => -> (t) {
    t[:ems_id].eq(nil).and(t[:storage_id].eq(nil))
  } do
    ems_id.nil? && storage_id.nil?
  end
  alias_method :archived, :archived?
```

Before 2
========

```ruby
class VmOrTemplate
  def normalized_state
    return self["normalized_state"] if has_attribute?("normalized_state")

    %w[archived orphaned template retired disconnected].each do |s|
      return s if send(:"#{s}?")
    end

    power_state&.downcase || "unknown"
  end

  virtual_attribute :normalized_state, :string, :arel => (lambda do |t|
    Arel::Nodes::Case.new
      .when(arel_table[:archived]).then(Arel.sql("'archived'"))
      .when(arel_table[:orphaned]).then(Arel.sql("'orphaned'"))
      .when(t[:template].eq(t.create_true)).then(Arel.sql("'template'"))
      .when(t[:retired].eq(t.create_true)).then(Arel.sql("'retired'"))
      .when(arel_table[:disconnected]).then(Arel.sql("'disconnected'"))
      .else(t.lower(t.coalesce([t[:power_state], Arel.sql("'unknown'")])))
  end)
end
```

After 2
=======

```ruby
class VmOrTemplate
  virtual_attribute :normalized_state, :string, :arel => -> (t) {
    Arel::Nodes::Case.new
      .when(arel_table[:archived]).then(Arel.sql("'archived'"))
      .when(arel_table[:orphaned]).then(Arel.sql("'orphaned'"))
      .when(t[:template].eq(t.create_true)).then(Arel.sql("'template'"))
      .when(t[:retired].eq(t.create_true)).then(Arel.sql("'retired'"))
      .when(arel_table[:disconnected]).then(Arel.sql("'disconnected'"))
      .else(t.lower(t.coalesce([t[:power_state], Arel.sql("'unknown'")])))
  } do
    %w[archived orphaned template retired disconnected].each do |s|
      return s if send(:"#{s}?")
    end

    power_state&.downcase || "unknown"
  end
end
```

kbrock:
WIP: still have to convert across
WIP: need to verify this doesn't bloat memory too much due to class closure in the lambda

Fixes #203 

===

Not thrilled that we are keeping the method closure from virtual_attribute. that has quite a bit of data in there and the lambda

### DIFF
--- a/lib/active_record/virtual_attributes.rb
+++ b/lib/active_record/virtual_attributes.rb
@@ -56,7 +56,7 @@ module ActiveRecord
         virtual_attribute(name, type, **options)
       end
 
-      def virtual_attribute(name, type, through: nil, uses: nil, arel: nil, source: name, default: nil, **options)
+      def virtual_attribute(name, type, through: nil, uses: nil, arel: nil, source: name, default: nil, **options, &block)
         name = name.to_s
         reload_schema_from_cache
 
@@ -77,6 +77,10 @@ module ActiveRecord
           #   Because we may not have loaded the class yet
           #   And we definitely have not loaded the database yet
           arel ||= virtual_delegate_arel(source, to_ref)
+        elsif block_given?
+          define_method(name) do
+            has_attribute?(name) ? self[name] : instance_eval(&block)
+          end
         end
 
         define_virtual_include(name, uses)
--- a/spec/db/models.rb
+++ b/spec/db/models.rb
@@ -61,32 +61,19 @@ class Author < VirtualTotalTestBase
   virtual_total :total_named_books, :named_books
   alias v_total_named_books total_named_books
 
-  def nick_or_name
-    has_attribute?("nick_or_name") ? self["nick_or_name"] : nickname || name
-  end
-
-  # sorry. no creativity on this one (just copied nick_or_name)
-  def name_no_group
-    has_attribute?("name_no_group") ? self["name_no_group"] : nickname || name
-  end
 
   # a (local) virtual_attribute without a uses, but with arel
   virtual_attribute :nick_or_name, :string, :arel => (lambda do |t|
-    t.grouping(Arel::Nodes::NamedFunction.new('COALESCE', [t[:nickname], t[:name]]))
-  end)
-
-  # We did not support arel returning something other than Grouping.
-  # this is here to test what happens when we do
-  virtual_attribute :name_no_group, :string, :arel => (lambda do |t|
     Arel::Nodes::NamedFunction.new('COALESCE', [t[:nickname], t[:name]])
-  end)
-
-  def first_book_name
-    has_attribute?("first_book_name") ? self["first_book_name"] : books.first.name
+  end) do
+   nickname || name
   end
 
-  def first_book_author_name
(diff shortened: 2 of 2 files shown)


lib/active_record/virtual_attributes.rb
```diff
@@ -77,6 +77,10 @@ def virtual_attribute(name, type, through: nil, uses: nil, arel: nil, source: na
           #   Because we may not have loaded the class yet
           #   And we definitely have not loaded the database yet
           arel ||= virtual_delegate_arel(source, to_ref)
+        elsif block_given?
+          define_method(name) do
+            has_attribute?(name) ? self[name] : instance_eval(&block)
```
kbrock, line 82: this is so nice.

We can also auto add a group or anything else that may change over time.
The current implementation doesn't allow us to do any common manipulation of the block return values

kbrock:
@Fryguy Where are you on this?
Could merge this and then not make the changes to manageiq/manageiq, or punt. Up to you.

1. This is fully backwards compatible.
2. The syntax is denser. (nice/bad)
3. The double block (one for arel and the other for ruby is not ideal.
4. Using `define_method` creates a block the side of the class for every virtual attribute.

Re 3: If you would like to pair up to derive SQL from Ruby, we can talk. It is very academic and probably not possible. I always think of [defunkt/ambition](https://github.com/defunkt/ambition), where he wrote LDAP/SQL from Ruby. Since we are standardized on the Prizm parser, it would have longer staying power.
This would sove the density problem.

Guessing improving `MiqExpression` would be a much better bang for your buck.

Re 4: This was the reason I aborted. Didn't profile this, but there is a reason why Rails pushes `:if => :method_symbol` and `:unless => :method_symbol` instead of blocks for most of their interfaces.

Fryguy:
I love this. Waiting on moving forward at the moment though because @jrafanie is doing Rails 8 support first and I don't want to get in his way.

